PR #3, opened 2021-10-25
Updated db_handler.py

mukhtarB:
- Added CRUD functionality for accessing the db.
- Converted read query to a more generic read.
- It uses the old plugin ID for dms (since we dont have a combined plugin ID yet. 
     Effect -> all data will be saved on dm collections)
- ORG_ID is also the one we've been using before
- Centrifugo token is older one (which i believe was specific for dms only also)
- login_user() is added temporarily, but it needs to go in the long run and become more dynamic

### DIFF
--- a/backend/utils/db_handler.py
+++ b/backend/utils/db_handler.py
@@ -4,27 +4,11 @@ import requests
 import json
 from datetime import datetime, timedelta
 
-
 from requests import exceptions
 
 
-def login_user():
-    data = {"email": "sam@gmail.com", "password": "Owhondah"}
-    try:
-        response = requests.post(url="https://api.zuri.chat/auth/login", json=data)
-    except requests.exceptions.RequestException as e:
-        return e
-    if response.status_code == 200:
-        return response.json()["data"]["user"]["token"]
-    else:
-        return None
-
-
-PLUGIN_ID = "61696380b2cc8a9af4833d80"
+PLUGIN_ID = ""
 ORG_ID = "61695d8bb2cc8a9af4833d46"
-ROOMS = "dm_rooms"
-MESSAGES = "dm_messages"
-header = {"Authorization": f"Bearer {login_user()}"}
 
 
 class DataStorage:
--- a/backend/utils/db_handler.py
+++ b/backend/utils/db_handler.py
@@ -22,7 +22,6 @@ def login_user():
 
 PLUGIN_ID = "61696380b2cc8a9af4833d80"
 ORG_ID = "61695d8bb2cc8a9af4833d46"
-CENTRIFUGO_TOKEN = "58c2400b-831d-411d-8fe8-31b6e337738b"
 ROOMS = "dm_rooms"
 MESSAGES = "dm_messages"
 header = {"Authorization": f"Bearer {login_user()}"}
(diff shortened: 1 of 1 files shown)


backend/utils/db_handler.py
```diff
@@ -0,0 +1,171 @@
...
+            response = requests.put(url=self.write_api, json=body)
+        except requests.exceptions.RequestException as e:
+            print(e)
+            return None
+        if response.status_code == 200:
+            return response.json()
+        else:
+            return {"status_code": response.status_code, "message": response.reason}
+
+    # async def read(self, collection_name, filter={}):
```
zxenonx: We can remove that part as it has been refactored below

backend/utils/db_handler.py
```diff
@@ -0,0 +1,171 @@
...
+        except requests.exceptions.RequestException as e:
+            print(e)
+            return None
+        if response.status_code == 200:
+            return response.json().get("data")
+        else:
+            return {"status_code": response.status_code, "message": response.reason}
+
+    async def delete(self, collection_name, document_id):
+    """
```
zxenonx, line 139: For the docstrings, i want us to stick with Google doctrings format:

"""Gets and prints the spreadsheet's header columns

Args:
    file_loc (str): The file location of the spreadsheet
    print_cols (bool): A flag used to print the columns to the console
        (default is False)

Returns:
    list: a list of strings representing the header columns
"""
From the Google Python Style Guide: https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings

zxenonx (changes requested):

backend/utils/db_handler.py
zxenonx: @mukhtarB

backend/utils/db_handler.py
zxenonx: @mukhtarB

backend/utils/db_handler.py
mukhtarB: Oh you mean the commented out code right... just left it incase I had to use it later... I'm definitely taking it out

backend/utils/db_handler.py
mukhtarB: Okay so args and returns, as opposed to params and returns.... 
Also, do you want to add a raises field @zxenonx

backend/utils/db_handler.py
mukhtarB: @zxenonx

zxenonx: Yeah, the commented out code. OK, please remove it ASAP. 👍

backend/utils/db_handler.py
zxenonx: Only if the function raises an exception

backend/utils/db_handler.py
mukhtarB: done

backend/utils/db_handler.py
mukhtarB: done

zxenonx (approved):

o4codes (approved):

PR #6, opened 2021-10-26
Centrifugo Handler 

o4codes:
This PR contains an Events whic contain an enum of all events to be sent through crntrifugo.
Therefore the publish function takes in it's parameters an event enum to which it publishes in it's data to to a room.

Also all data sent to centrifugo is unified in the centrifugo function there only required data is to be input as data in the parameter.

A standard response is provided as a return to the publish and unpublish function. That response can be used as response to the endpoints call for standardised response


mukhtarB (approved):

backend/utils/centrifugo_handler.py
```diff
@@ -0,0 +1,139 @@
...
+from typing import Any, Dict, Optional
+from requests.exceptions import RequestException
+import requests
+
+
+CENTRIFUGO_HOST = "https://realtime.zuri.chat/api"
+CENTRIFUGO_API_TOKEN = "58c2400b-831d-411d-8fe8-31b6e337738b"
+
+
+class Events(str, Enum):
```
zxenonx, line 11: Take care of this events class @o4codes .

zxenonx (approved):

zxenonx:
@o4codes , please add description to the PR

PR #8, opened 2021-10-29
startup.sh script refactored

o4codes:
startup.sh script has been refactored to use both venv and pipenv

to use venv 
sh startup.sh venv

to use pipenv
sh startup.sh pipenv

@mukhtarB 
@zxenonx 
@Delight-Fela-Steve 
@man-of-mind

### DIFF
--- a/backend/readme.md
+++ b/backend/readme.md
@@ -1,0 +1,3 @@
+# Contributions Guidelines
+
+Contribution Guidelines for the Backend will be updated soon
--- a/backend/startup.sh
+++ b/backend/startup.sh
@@ -1,12 +1,39 @@
-pip install -r requirements.txt
-cd -
-if [[ -f ./.git/hooks/pre-commit ]]
+
+if [ $1 = "venv" ] || [ $1 = "pipenv" ]
 then
-    echo "pre-commit already setup"
-    cd backend
+
+    if [ -z "$VIRTUAL_ENV" ]
+    then
+        echo "No VIRTUAL_ENV set"
+    else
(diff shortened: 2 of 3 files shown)


man-of-mind (approved):
Well done boss, what happens when u already have a virtual environment set up with a different name aside venv on a linux?
If u create a new virtual environment, will everything be okay? I mean no conflict between the two virtual environments.

o4codes:
The venv is just a name.
If you don't have any virtual environment activated the script will not proceed

o4codes:
Further improvements will be made to search for existing virtual environment in the folder and use them

man-of-mind:
Oh....okay, nice
Well done boss.

zxenonx:
> Well done boss, what happens when u already have a virtual environment set up with a different name aside venv on a linux?
> If u create a new virtual environment, will everything be okay? I mean no conflict between the two virtual environments.

@man-of-mind, I think in ```
sh startup.sh venv
```, venv is actually a name. Your env folder could be backend-env or env, it doesn't matter, because the name of your virtual env folder is not used in the script. @o4codes, right?

zxenonx (approved):

man-of-mind:
@o4codes @zxenonx I don't think there will be any problem even with existing virtual environment, a new environment should be created (venv) and then start it, so far it's the venv that is started and not the other one.

o4codes:
Well.
For future improvement the script will be made to check for any virtual environment in the present working directory and use that virtual environment.

man-of-mind:
Well done boss @o4codes 
U be baba

PR #10, opened 2021-10-30
Centrifugo test

o4codes:
1.Created tests for the centrifugo handler.
2. Requests to centrifugo is changed to use httpx in place of requests, in order to support asynchronous requests.

Kindle review
@mukhtarB 
@zxenonx 
@man-of-mind 
@Delight-Fela-Steve 
@austinoski


Delight-Fela-Steve (approved):

man-of-mind (commented):
Boss, i was thinking we will be mocking the responses from 3rd party api call so we don't load the server with fake data.

The data won't persist, this makes sense, well done sir.

backend/utils/centrifugo_handler.py
```diff
@@ -1,14 +1,13 @@
...
-import requests
 
+import httpx
 
 CENTRIFUGO_HOST = "https://realtime.zuri.chat/api"
 CENTRIFUGO_API_TOKEN = "58c2400b-831d-411d-8fe8-31b6e337738b"
 
 
-class Events(str, Enum):
+class Events(Enum):
```
zxenonx, line 10: It's working without the str parameter?
@o4codes

mukhtarB (approved):

man-of-mind (approved):

zxenonx (approved):

backend/utils/centrifugo_handler.py
o4codes: Yes it works

PR #19, opened 2021-11-24
Fix staticfiles directory doesn't not exist error

zxenonx:
The error occurred as there is no frontend/dist folder created yet and it prevents the server to run.

- Solved by setting check_dir parameter to False.
- There's still a RuntimeError as the dist folder is still not created but **the server can run now**.
(RuntimeError: StaticFiles directory '../frontend/dist' does not exist.)


o4codes (approved):
Good work

mikenrowland (approved):
Good to go Chief.

PR #27, opened 2021-11-28
minor fixes to sidebar

Bwise1:
fixes made from the server

### DIFF
--- a/backend/endpoints/sync.py
+++ b/backend/endpoints/sync.py
@@ -60,29 +60,29 @@ async def dm_install(payload: InstallPayload):
     if channel_res is None:
         raise HTTPException(
             status_code=status.HTTP_424_FAILED_DEPENDENCY,
-            detail="unable to remove room member",
+            detail="unable to create default Channel",
         )
-    if isinstance(channel_res, dict):
-        raise HTTPException(
-            status_code=status.HTTP_424_FAILED_DEPENDENCY,
-            detail="unable to remove room member",
-        )
+#    if isinstance(channel_res, dict):
+#        raise HTTPException(
+#            status_code=status.HTTP_424_FAILED_DEPENDENCY,
+#            detail="unable to remove room member",
+#        )
 
     dm_res = await DB.write(ROOM_COLLECTION, data=dm)
 
     if dm_res is None:
         raise HTTPException(
             status_code=status.HTTP_424_FAILED_DEPENDENCY,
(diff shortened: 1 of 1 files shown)


o4codes (commented):
@Bwise1 
Good work
Please kindly read the reviews below.

backend/endpoints/sync.py
```diff
@@ -108,22 +108,22 @@ async def get_sidebar(org_id: str, member_id: str, room_type: RoomType):
...
-            status_code=status.HTTP_400_BAD_REQUEST, detail="room_type is required"
+            detail="user is not a valid object id",
         )
+   # if room_type is None:
+   #     raise HTTPException(
+   #         status_code=status.HTTP_400_BAD_REQUEST, detail="room_type is required"
+   #     )
 
-    data = await sidebar.format_data(org_id, member_id, room_type)
+    data = await sidebar.format_data(org, user, RoomType.CHANNEL)
```
o4codes, line 126: @Bwise1 
This line of code means only data for Channels is going to be recovered.
What about data for DM, because the former code took care of that ?

backend/endpoints/sync.py
```diff
@@ -94,7 +94,7 @@ async def dm_install(payload: InstallPayload):
         400: {"detail": "Object id is not valid"},
     },
 )
-async def get_sidebar(org_id: str, member_id: str, room_type: RoomType):
+async def get_sidebar(org: str, user: str):
```
o4codes, line 97: @Bwise1 
Since the repository handles for both DM and Channels, without the roomtype as a query parameter it's going to be difficult to include what roomtype data you're requesting for to be rendered on the sidebar

backend/endpoints/sync.py
almaudoh: The solution to this is to do two calls, one with `RoomType.CHANNEL` and one with `RoomType.DM`. We will have to discuss how this payload is structured.

almaudoh: Alternatively, we can refactor `sidebar.format_data()` to just give the complete payload.

backend/endpoints/sync.py
almaudoh: We can ask zuri main to provide `room_type`, but that would not be generalizable to other plugins. A better approach is to provide sidebar info for both at once. However, this needs to be discussed with respect to the payload structure for sidebar.

PR #28, opened 2021-11-29
Feat/setup-routing-zc-messaging

odelolajosh:
Set up routing


bori7 (approved):
Fine

TheIsrael1 (approved):

frontend/zc_messaging/package.json
```diff
@@ -53,6 +53,8 @@
     "react-icons": "^4.3.1",
     "react-redux": "^7.2.6",
     "react-router-dom": "^5.3.0",
-    "single-spa-react": "^4.3.1"
+    "redux": "^4.1.2",
```
tundealabi: the redux library is not needed, we already have @redux/toolkit installed

frontend/zc_messaging/src/App.js
```diff
@@ -1,17 +1,11 @@
...
 
 export default function App() {
   return (
-    <Switch>
-      <Route exact path="/all-dms" component={AllDms} />
-      <Route exact path="*">
-        <Redirect to="/all-dms" />
-      </Route>
-    </Switch>
+    <Routes />
```
tundealabi: am not sure we should have this in a component, since App is our entry point then the route definitions should be here instead. or is there a reason for this

frontend/zc_messaging/src/component/ChannelDetails/index.jsx
```diff
@@ -0,0 +1,10 @@
+import React from "react"
```
tundealabi: for consistency, can we have the folder names in lowercase - "channel_details"

frontend/zc_messaging/src/pages/DmBrowser/index.jsx
```diff
@@ -0,0 +1,10 @@
+import React from "react"
```
tundealabi: is dm browser not the same as "all-dms" which has already been initialized

frontend/zc_messaging/src/pages/MessageBoard/index.jsx
```diff
@@ -0,0 +1,29 @@
+import React from "react"
```
tundealabi: i thought message board is a component from zuri UI, why do we have it as a page here cos we don't have any page with /messageboard

frontend/zc_messaging/src/routes/index.jsx
```diff
@@ -0,0 +1,38 @@
+import React from "react"
```
tundealabi, line 1: this idea of having a route folder is kinda strange to me, haven't seen this in react projects. Is there a reason 🤔

frontend/zc_messaging/package.json
odelolajosh: It was installed during the initial project setup. I'll remove it.

frontend/zc_messaging/src/component/ChannelDetails/index.jsx
odelolajosh: Sure. I actually prefer that

frontend/zc_messaging/src/routes/index.jsx
odelolajosh: The route folder holds components relating to routing. Currently, there are two components for routing.

It makes the App component a lot neater, separating routing logic from other logics to be implemented in the App component

frontend/zc_messaging/src/pages/MessageBoard/index.jsx
odelolajosh: The MessageBoard component is the page that deals with messaging for the plugin. It should house the ZuriUi MessageBoard component. It wouldn't be okay to use the ZuriUi MessageBoard component as the messaging page directly.

frontend/zc_messaging/src/pages/DmBrowser/index.jsx
odelolajosh: I had it before you added all-dms folder.

I didn't remove it because I preferred the name **dm-browser** to **all-dms**. What do you think?

frontend/zc_messaging/src/App.js
odelolajosh: The reason was just to separate routing logic form other business logic from App.js

It's basically the same thing

tundealabi: i know it's the same thing but that means the app.js file might only contain that "route" component

frontend/zc_messaging/src/pages/DmBrowser/index.jsx
tundealabi: okay... that means we will move the content of "all-dms" into "dm-browser"

frontend/zc_messaging/src/pages/MessageBoard/index.jsx
tundealabi: Yeah, we can't just have the component from zuri UI like that. so I was thinking of having something like this

DM - /dm/:id
Channel - /ch/:id
for separation of concerns ... and both pages can just share components including the zuri UI component

frontend/zc_messaging/src/routes/index.jsx
tundealabi: okay

frontend/zc_messaging/src/App.js
odelolajosh: There is another route file for right sidebar components like thread, channel details, member profile

tundealabi: okay

tundealabi (approved):
Good to go @zxenonx

PR #29, opened 2021-11-30
Add to room endpoint

mikenrowland:
An endpoint that allows for a member or multiple members to be added to any room of type Channels or Group DM.


backend/endpoints/rooms.py
```diff
@@ -1,9 +1,11 @@
 from fastapi import APIRouter, BackgroundTasks, HTTPException, status
 from fastapi.responses import JSONResponse
 from schema.response import ResponseModel
-from schema.room import Room, RoomRequest
+from schema.room import AddToRoom, Role, Room, RoomRequest, RoomType
+from starlette.responses import Response
```
smyja: Can imports like this be added to one file so everyone can use it? That way the number of code lines will 
be reduced and the codebase will follow the DRY pattern (Do not repeat yourself) 

@zxenonx

mikenrowland: I don't know how that can be resolved, most of them are fastapi inbuilt imports

backend/endpoints/rooms.py
```diff
@@ -61,3 +63,69 @@ async def create_room(
...
+                        raise HTTPException(
+                            detail="the max number for a Group_DM is 9",
+                            status_code=400,
+                        )
+                    new_data = {person: new_member.get(str(person))}
+                    room_members.update(new_data)
+
+            update_members = {"room_members": room_members}
+            update_res = await DB.update(
+                "chat_rooms", document_id=room_id, data=update_members
```
o4codes: I think DB.update("chat_rooms"......)
can be replaced with DB.update(ROOM_COLLECTION......)
where ROOM_COLLECTION is the variable in room_util.py that holds to name of room collection

backend/endpoints/rooms.py
```diff
@@ -61,3 +63,69 @@ async def create_room(
...
+            )
+            if update_res and update_res.get("status_code", None) is None:
```
o4codes: Centrifugo functions can be put in a background task so that they do not affect the endpoint response.
Because centrifugo maybe down, but in essence the data should be saved in the core.
Therefore, centrifugo being up or down shouldn't affect the response, as long as the data is saved to core, it should return a success response, and the call to centrifugo shouldn't block the request

backend/schema/room.py
```diff
@@ -231,3 +231,9 @@ def is_object_id(cls, values):
             )
 
         return values
+
+
+class AddToRoom(BaseModel):
```
o4codes, line 237: Nice work.
Do you think it's possible in fastapi it's possible to add a body parameter in the endpoint function without creating a class for it, so that rather than creating a class with a single instance variable would be avoided ?

backend/endpoints/rooms.py
```diff
@@ -61,3 +63,69 @@ async def create_room(
...
+
+    Returns:
+        HTTP_200_OK: member added
+    Raises:
+        HTTP_400_BAD_REQUEST: the max number for a Group_DM is 9
+        HTTP_401_UNAUTHORIZED: member not an admin
+        HTTP_403_FORBIDDEN: DM room or not found
+    """
+
+    new_member = data.dict().get("new_member")
```
o4codes, line 119: Good work.
What if we created a add_to_room function in the room_utils where most of this codes are placed in and also this would allow the function to be used any where, so we therefore reference the function inside this endpoint ?

o4codes (changes requested):
Good work @mykie88 . I have made some reviews,

backend/endpoints/rooms.py
```diff
@@ -61,3 +63,69 @@ async def create_room(
...
+    room = await get_room(org_id=org_id, room_id=room_id)
+
+    if room is not None and room["room_type"] != RoomType.DM:
```
o4codes: What do you think if the nestings were reduced and nested ifs were avoided. This isn't consequential though, but might make the code easily readable.

Something like
`
if room is None and room['room_type'] == RoomType.DM:
      raise HTTPException
if member['role'] != Role.ADMIN:
       raise HTTPException

`

backend/endpoints/rooms.py
mikenrowland: okay

backend/endpoints/rooms.py
mikenrowland: hmmm, I saw what you did there. I think I agree too. I will see how I can make that work.

backend/endpoints/rooms.py
mikenrowland: That sounds great, but I don't see this function reusable anywhere else in the code base. The purpose is unilateral and all possible use cases have been factored in.

backend/schema/room.py
mikenrowland: yeah absolutely I think that is possible. I will effect the change.

backend/endpoints/rooms.py
mikenrowland: I agree. I will see to that too. Thanks

backend/endpoints/rooms.py
o4codes: Yeah, I have not seen other general purpose for it too though.
I just felt separating the actual business logic code from the controller logic, might help in re-usability.
But also this is also is not consequential for now as the purpose is unilateral.

backend/schema/room.py
mikenrowland: After trying different ways to add the input data as a body param, IMO using the class was the best way to keep the code clean and the function less complex.

o4codes: Alright thanks

backend/endpoints/rooms.py
```diff
@@ -62,3 +63,77 @@ async def create_room(
...
+    if member is None or member["role"] != Role.ADMIN:
+        raise HTTPException(
+            status_code=401, detail="member not in room or not an admin"
+        )
+
+    if room["room_type"] == RoomType.CHANNEL:
+        room["room_members"].update(new_member)
+
+    if room["room_type"] == RoomType.GROUP_DM:
+        for person in list(new_member.keys()):
```
o4codes: Can we avoid this for loop and just do instead
`
room['room_members'].update(new_data) # this will update the new room with the new members

if len(room["room_members"].keys()) >= 9:
                raise HTTPException(
                    detail="the max number for a Group_DM is 9",
                    status_code=400,
                ) # this then the checks the total members of people in the room including those added
`

mikenrowland: the loop was put in place in case of bulk adding of members to a group dm so that the count will prevent the update when the len(room["room_members"].keys()) is > 9 at a particular iteration

mikenrowland: oh! you mean update the members list before I check len(room["room_members"].keys()) right?

mikenrowland: done! works great

o4codes (approved):
Good work.
Thanks for effecting the reviews @mykie88

o4codes:
@zxenonx please review

backend/endpoints/rooms.py
```diff
@@ -62,3 +63,75 @@ async def create_room(
...
+    member_id: str,
+    background_tasks: BackgroundTasks,
+):
+    """Adds a new member(s) to a room
+    Args:
+        data: a pydantic schema that defines the request params
+        org_id (str): A unique identifier of an organisation
+        room_id: A unique identifier of the room to be updated
+        member_id: A unique identifier of the member initiating the request
+
```
zxenonx, line 101: Please, kindly update the docstring.
The parameter `background_tasks` is missing here.

backend/endpoints/rooms.py
zxenonx: @mykie88, please do so. Actually, we can extract out some smaller functions from here.

backend/endpoints/rooms.py
```diff
@@ -62,3 +63,75 @@ async def create_room(
...
+    if room["room_type"] == RoomType.GROUP_DM:
+        room["room_members"].update(new_member)
+        if len(room["room_members"].keys()) > 9:
+            raise HTTPException(
+                detail="the max number for a Group_DM is 9",
+                status_code=400,
+            )
+
+    update_members = {"room_members": room["room_members"]}
+    update_res = await DB.update(
```
zxenonx: Here i suggest we write 
```suggestion
    update_response = await DB.update(
```
OR
```suggestion
    response = await DB.update(
```

backend/endpoints/rooms.py
```diff
@@ -62,3 +63,75 @@ async def create_room(
...
+
+    background_tasks.add_task(
+        centrifugo_client.publish,
+        room=room_id,
+        event=Events.ROOM_MEMBER_ADD,
+        data=new_member,
+    )
+
+    if update_res and update_res.get("status_code", None) is None:
+        return JSONResponse(content=update_res, status_code=200)
```
zxenonx: Here too: `update_response` or `response`

backend/endpoints/rooms.py
```diff
@@ -62,3 +63,75 @@ async def create_room(
...
+        HTTP_403_FORBIDDEN: DM room or not found
+    """
+
+    DB = DataStorage(org_id)
+    new_member = data.dict().get("new_member")
+    room = await get_room(org_id=org_id, room_id=room_id)
+    member = room.get("room_members").get(str(member_id))
+
+    if room is None or room["room_type"] == RoomType.DM:
+        raise HTTPException(status_code=403, detail="DM room or not found")
```
zxenonx: In the code, can we replace the numeric value of the status code with the convenience variables from` fastapi.status`. I think it will be more understandable.
What do you think?
```suggestion
        raise HTTPException(status_code=HTTP_403_FORBIDDEN, detail="DM room or not found")
```

backend/endpoints/rooms.py
```diff
@@ -62,3 +63,75 @@ async def create_room(
...
+
+    if member is None or member["role"] != Role.ADMIN:
+        raise HTTPException(
```
zxenonx, line 130: Same here too. status_code variable

backend/endpoints/rooms.py
```diff
@@ -62,3 +63,75 @@ async def create_room(
...
+
+    if room["room_type"] == RoomType.CHANNEL:
+        room["room_members"].update(new_member)
+
+    if room["room_type"] == RoomType.GROUP_DM:
+        room["room_members"].update(new_member)
+        if len(room["room_members"].keys()) > 9:
+            raise HTTPException(
+                detail="the max number for a Group_DM is 9",
+                status_code=400,
```
zxenonx: Same here too. status_code variable

backend/utils/room_utils.py
```diff
@@ -58,7 +58,7 @@ async def get_room(org_id: str, room_id: str) -> dict:
 
     if response and "status_code" not in response:
         return response
-    return {}
+    return None
```
zxenonx: Have u faced an issue with the function returning an empty dict if there is no room found?

zxenonx (changes requested):
Thanks, @mykie88 for the great work.
Just left a few comments.

backend/endpoints/rooms.py
mikenrowland: Okay, I was just trying to keep the characters in each line of code as few as possible

backend/utils/room_utils.py
mikenrowland: yes I did, an empty dict is not a none value, so when a room is not returned the error is not caught and the code fails

backend/endpoints/rooms.py
mikenrowland: Okay, I will update that

backend/utils/room_utils.py
zxenonx: I see. But empty dicts evaluate to False.
Therefore, `if not room` could have worked.
@mykie88

mikenrowland: yes it would have, but this seems like a matter of preference, and quite inconsequential. How does it affect anything?

zxenonx: {} vs None battle loading...

mikenrowland: Lmao... let me just be the bigger person

backend/endpoints/rooms.py
```diff
@@ -62,3 +63,95 @@ async def create_room(
...
+            detail="member not in room or not an admin",
+        )
+
+    if room["room_type"] == RoomType.CHANNEL:
+        room["room_members"].update(new_member)
+
+    if room["room_type"] == RoomType.GROUP_DM:
+        room["room_members"].update(new_member)
+        if len(room["room_members"].keys()) > 9:
+            raise HTTPException(
```
almaudoh, line 141: Would it not be better to do the test and throw this exception before adding the member to the room? Doesn't `update` persist the new member already?

i.e. `if len(room["room_members"].keys()) >= 9:` then throw the exception first.

Pipfile.lock
```diff
@@ -0,0 +1,793 @@
+{
```
almaudoh, line 1: Should this change be in the PR?

backend/endpoints/rooms.py
o4codes: The proces @mykie88 is following is that he's updating the members_list.
Then doing the test to confirm that even with the updated members the count doesn't still exceed 9

Then if the test passes, he persist the data.

mikenrowland: no @Aniebet that update is just a dict method. The data is persisted further down the code at the DB.update method of the datastorage class

zxenonx: @almaudoh

PR #30, opened 2021-11-30
Fixed messaging install endpoint. Updated sidebar endpoint.

almaudoh:
- Fixed messaging install endpoint
- Updated sidebar endpoint


zxenonx (approved):
LGTM

PR #31, opened 2021-12-02
Fix get sidebar issue

zxenonx:
The function returned: 

-  a Key value error on this line [L36](https://github.com/zurichat/zc_messaging/blob/edf909c927ed8caa41a5585212c76d20b3b48a3b/backend/utils/sidebar.py#L36)
- Response.model was waiting for a dict : [L115](https://github.com/zurichat/zc_messaging/blob/edf909c927ed8caa41a5585212c76d20b3b48a3b/backend/endpoints/sync.py#L115)

Fixes: 

- a default value for the pop() function
- a dict object for Response.model


o4codes (approved):

backend/endpoints/sync.py
```diff
@@ -112,5 +112,6 @@ async def get_sidebar(org: str, user: str):
     channel_data = await sidebar.format_data(org, user, RoomType.CHANNEL)
     dm_data = await sidebar.format_data(org, user, RoomType.DM)
     return JSONResponse(
-        content=ResponseModel.success([channel_data, dm_data]), status_code=status.HTTP_200_OK
+        content=ResponseModel.success({"channels": channel_data, "dms": dm_data}),
```
almaudoh: How can we make this return a list instead of a dict?

zxenonx: @almaudoh, fixed. It accepts a list now.

PR #36, opened 2021-12-04
Feat/message board

tundealabi:
- redirect to the messaging board page by default
- fetch a default room if none is passed
- import and use message board from zuri UI

### DIFF
--- a/frontend/root-config/src/index.ejs
+++ b/frontend/root-config/src/index.ejs
@@ -53,7 +53,7 @@
         "imports": {
           "single-spa": "https://cdn.jsdelivr.net/npm/single-spa@5.9.0/lib/system/single-spa.min.js",
           "vue": "https://cdn.jsdelivr.net/npm/vue@2.6.11/dist/vue.min.js",
-          "vue-router": "https://cdn.jsdelivr.net/npm/vue-router@3.1.6/dist/vue-router.min.js"
+          "vue-router": "https://cdn.jsdelivr.net/npm/vue-router@3.1.6/dist/vue-router.min.js",
           "@zuri/zuri-ui": "https://staging.zuri.chat/zuri-zuri-ui.js"
         }
       }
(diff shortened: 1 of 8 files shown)


frontend/root-config/src/index.ejs
```diff
@@ -53,7 +53,8 @@
         "imports": {
           "single-spa": "https://cdn.jsdelivr.net/npm/single-spa@5.9.0/lib/system/single-spa.min.js",
           "vue": "https://cdn.jsdelivr.net/npm/vue@2.6.11/dist/vue.min.js",
-          "vue-router": "https://cdn.jsdelivr.net/npm/vue-router@3.1.6/dist/vue-router.min.js"
+          "vue-router": "https://cdn.jsdelivr.net/npm/vue-router@3.1.6/dist/vue-router.min.js",
```
zxenonx, line 56: @tundealabi, please why are we importing vue and vue-router?

bori7: Same thoughts exactly.

Isn't this SPA a react app?

frontend/zc_messaging/src/utils/fetchDefaultRoom.js
```diff
@@ -0,0 +1,16 @@
...
+const fetchDefaultRoom = async (orgId, userId) => {
+  const response = await axios.get(
+    `https://chat.zuri.chat/api/v1/sidebar?org=${orgId}&user=${userId}`
+  )
+  const { public_rooms } =
+    response.data?.data?.find(room => room.name === "Channels") || {}
+  const defaultRoom =
+    public_rooms?.find(room => room.room_name === "general") || {}
+  return {
+    room_id: defaultRoom.room_id
```
bori7, line 12: The case for defaultRoom if its null, doesn't have a room_id-> {}

frontend/root-config/src/index.ejs
tundealabi: dunno.. didn't add that

frontend/zc_messaging/src/utils/fetchDefaultRoom.js
tundealabi: won't throw an error, will just be undefined

frontend/root-config/src/index.ejs
zxenonx: I think it was @benjamin-bala, maybe for when a Vue app will be integrated...

PR #37, opened 2021-12-05
Refactor

o4codes:
Made some few refactorings.
1. Validated all DM, GROUP DM rooms are set to private
2. Removed the install class and replaced it with request body directly in the function parameters because the class will not be reusable.
3. Validated organisation_id and user_id in install request body are valid object_id

### DIFF
--- a/backend/endpoints/sync.py
+++ b/backend/endpoints/sync.py
@@ -12,7 +12,16 @@ from utils.sidebar import sidebar
 router = APIRouter()
 
 
-@router.post("/install")
+@router.post(
+    "/install",
+    status_code=status.HTTP_200_OK,
+    responses={
+        200: {"detail": "installtion successful"},
+        400: {"detail": "Object id is not valid"},
+        422: {"detail": "Field required"},
+        424: {"detail": "Instaalltion Failed"},
+    },
+)
 async def dm_install(organisation_id: str = Body(...), user_id: str = Body(...)):
     """This endpoint is called when an organisation wants to install the
     DM plugin for their workspace.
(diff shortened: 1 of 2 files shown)


backend/endpoints/sync.py
```diff
@@ -13,47 +12,64 @@
...
+        )
 
-    DB = DataStorage(payload.organisation_id)
+    DB = DataStorage(organisation_id)
     channel = {
         "room_name": "general",
         "room_type": RoomType.CHANNEL,
         "room_members": {
-            payload.user_id: {"role": Role.ADMIN, "starred": False, "closed": False}
+            user_id: {"role": Role.ADMIN, "starred": False, "closed": False}
```
mikenrowland, line 58: Seen. I've tried this though, not exactly the same way but something similar and the "Role.ADMIN" part was throwing an error when I tried to update the room using DB.update.
But I will still check it and see if it solves the problem. Thanks

backend/schema/room.py
```diff
@@ -138,6 +138,7 @@ def validate_dm(cls, values):
                     detail="DM or Group DM should not have a description",
                 )
 
+        values["is_private"] = True
```
mikenrowland: Is this supposed to be outside of the first 'IF' block at line 83? because I think this will end up setting "is_private" field for Channels to TRUE too.
@o4codes

mikenrowland (commented):
Well done Chief I left some comments.

backend/endpoints/sync.py
```diff
@@ -13,47 +12,64 @@
 router = APIRouter()
 
 
-class InstallPayload(BaseModel):
-    """Installation payload model."""
+@router.post("/install")
```
zxenonx: Please, kindly complete with `status_code` and `responses`

backend/endpoints/sync.py
```diff
@@ -13,47 +12,64 @@
...
+async def dm_install(organisation_id: str = Body(...), user_id: str = Body(...)):
```
zxenonx, line 25: 👍

backend/schema/room.py
zxenonx: Yes, indeed. I think this is a mistake @o4codes

zxenonx (changes requested):

zxenonx:
Thanks for the refactoring @o4codes

backend/endpoints/sync.py
o4codes: Alright

backend/endpoints/sync.py
o4codes: Thanks Fixed

backend/schema/room.py
o4codes: Thanks, Fixed

o4codes:
@zxenonx 
@mykie88 

Thanks for the reviews.
The reviews have been effected

zxenonx (approved):
Thanks @o4codes  for the changes.
LGTM.

FlorenceEgwu:
Looks good to me.
Well than @o4codes .

PR #39, opened 2021-12-08
Refactor

o4codes:
checks if room creator is in room_members
if not adds the room_creator as an admin in the room_members of the room, in the create_room endpoint


backend/endpoints/rooms.py
```diff
@@ -41,6 +41,16 @@ async def create_room(
 
     DB = DataStorage(org_id)
     room_obj = Room(**request.dict(), org_id=org_id, created_by=member_id)
+
+    if (
```
zxenonx: Do we need parentheses here? @o4codes

zxenonx (commented):
Thanks for the work.

backend/endpoints/rooms.py
o4codes: @zxenonx 
It's been removed now.
Thanks

zxenonx (approved):
LGTM.

PR #43, opened 2021-12-11
Added Send_message endpoint 

FlorenceEgwu:
Added the send message endpoint.
Fixed minor bugs in rooms.py
Created a test for a successful and unsuccessful send_message operation


backend/endpoints/messages.py
```diff
@@ -1,3 +1,78 @@
...
+from schema.response import ResponseModel
+from starlette.responses import JSONResponse
+from utils.centrifugo import Events, centrifugo_client
+from utils.db import DataStorage
+from utils.room_utils import get_room
 
 router = APIRouter()
+
+MESSAGE_COLLECTION = "chat_messages"
+test_room = "61b3fb328f945e698c7eb396"
```
o4codes: is the test_room variable still needed

backend/endpoints/messages.py
```diff
@@ -1,3 +1,78 @@
...
+
+    if response.get("status_code") != 201:
+        raise HTTPException(
+            status_code=status.HTTP_424_FAILED_DEPENDENCY,
+            detail={"Message not sent": response},
+        )
+    message_obj.message_id = response["data"]["object_id"]
+
+    background_tasks.add_task(
+        centrifugo_client.publish, room_id, Events.MESSAGE_CREATE,
```
o4codes: The centrifugo_client.publish method requires an extra parameter which is "data", 
data is a parameter that refers to centrifugo data being published so that it can be received and rendered on the frontend.

in this case scenario data would be a dictionary  that has the following: 
`{
"room_id":"id of the room being published to",
"message_id":"id of the message sent"
"content":"text content of message sent",
"sender_id":"id of the sender who sent the message",
"created_at":"datetime of message sent"
}
`

backend/endpoints/messages.py
```diff
@@ -1,3 +1,78 @@
...
+        HTTPException [404]: Sender not in room
+        HTTPException [404]: Room does not exist
+        HTTPException [424]: "message not sent"
+    """
+    DB = DataStorage(org_id)
+    message_obj = Message(**request.dict(), org_id=org_id, room_id=room_id,
+                            sender_id= sender_id)
+    response = await DB.write(MESSAGE_COLLECTION, message_obj.dict())
+
+    if response.get("status_code") != 201:
```
o4codes: The response variable is going to be the return value of the db.write operation.
The only time it will have the key  "status_code" is actually when the request fails.

Though this works but request.get("status_code") can never be equal to 201 because db.write isn't returning the response rather it is returning response.json() on a success request. The presence of key "status_code" actually means the request fails.

therefore the condition can be rewritten as "if response and  response.get("status_code") == None"
this condition denotes a success, based on what db.write() is returning

backend/schema/message.py
```diff
@@ -11,22 +13,58 @@ class Reaction(BaseModel):
...
 
+class Thread(ThreadRequest):
+    sender_id: str
+    room_id: str
+    org_id:str
+    message_id:str
+
+    @root_validator(pre=True)
+    @classmethod
+    def validates_channels(cls, values):
```
o4codes: Can this function be renamed to "validate_message"

backend/tests/endpoints/test_message.py
```diff
@@ -2,3 +2,45 @@
...
+
+def test_send_message():
+    """ test for the send message endpoint"""
+
+    fake_response_output = {
+        "message_id": "33090940900",
+        "message_data": fake_message_data
+    }
+
+    response = client.post(
```
o4codes, line 29: Nice work.
The only issue is that the test_send_message would actually be populating the db with test data.
This can be avoided if the actually write request was a mock function.
Or 
A test data can be sent. 
On validating that it was actually sent, a delete request is sent to the delete the message. so that the db is not populated. The issue with this is that your time taken for the tests to be carried would be longer.

Therefore, the best approach would be to mock the db.write function.

backend/utils/db.py
```diff
@@ -56,7 +56,7 @@ async def write(self, collection_name, data):
         if response.status_code == 201:
             return response.json()
 
-        return {"status_code": response.status_code, "message": response.reason}
+        return {"status_code": response.status_code, "message": response.json()}
```
o4codes, line 59: Good work

o4codes (changes requested):
NIce work @AI-fae 
Please kindly go through the reviews

backend/endpoints/messages.py
```diff
@@ -1,3 +1,78 @@
 from fastapi import APIRouter
+from fastapi import APIRouter
+from fastapi import APIRouter, BackgroundTasks, HTTPException, status
+from schema.message import Message, ThreadRequest
+from schema.response import ResponseModel
+from starlette.responses import JSONResponse
+from utils.centrifugo import Events, centrifugo_client
+from utils.db import DataStorage
+from utils.room_utils import get_room
```
zxenonx: `get_room` imported but unused. You can remove it.
`git commit `would not even pass with the pre-commit config. @AI-fae 😒😒

backend/schema/message.py
```diff
@@ -1,8 +1,10 @@
+import asyncio
+import concurrent.futures
 from datetime import datetime
 from typing import List
-
-from pydantic import AnyHttpUrl, BaseModel
-
+from fastapi import HTTPException, status
+from pydantic import AnyHttpUrl, BaseModel, Field, root_validator
```
zxenonx, line 6: `Field` imported but unused. You can remove it.
git commit would not even pass with the pre-commit config. @AI-fae 😒

backend/schema/message.py
```diff
@@ -11,22 +13,58 @@ class Reaction(BaseModel):
...
     text: str
     reactions: List[Reaction] = []
     files: List[AnyHttpUrl] = []
     saved_by: List[str] = []
-    is_pinned: bool = False
-    is_edited: bool = False
     created_at: str = str(datetime.utcnow())
 
 
+class Thread(ThreadRequest):
```
zxenonx: Docstring is missing.
Kindly update the class with a docstring.
`git commit `would not pass with the pre-commit config. @AI-fae

backend/endpoints/messages.py
zxenonx: @AI-fae, you didn't use it here, right?

backend/endpoints/messages.py
```diff
@@ -1,3 +1,78 @@
...
+    },
+)
+async def send_message(
+    org_id, room_id, sender_id, request:ThreadRequest, background_tasks: BackgroundTasks
+):
+    """Creates and sends a message from one user to another.
+    Registers a new document to the chats database collection.
+    Returns the message info and document id if message is successfully created
+    while publishing to all members of the room in the background
+    Args:
```
zxenonx, line 30: `sender_id` and `background_tasks` are missing in the Args section of the docstring.

backend/endpoints/messages.py
```diff
@@ -1,3 +1,78 @@
...
+    "/org/{org_id}/rooms/{room_id}/sender/{sender_id}/messages",
+    response_model=ResponseModel,
+    status_code=status.HTTP_201_CREATED,
+    responses={
+        404: {"detail": "room or sender not found"},
+        424: {"detail": "ZC Core Failed"},
+    },
+)
+async def send_message(
+    org_id, room_id, sender_id, request:ThreadRequest, background_tasks: BackgroundTasks
```
zxenonx: As this function sends messages, i think it will make more sense to rename the request schema into `MessageRequest`

backend/endpoints/messages.py
FlorenceEgwu: oh my! forgot to take it after testing.

backend/endpoints/messages.py
```diff
@@ -1,3 +1,78 @@
...
+                "thread": [],
+                        }
+            }
+    Raises:
+        HTTPException [404]: Sender not in room
+        HTTPException [404]: Room does not exist
+        HTTPException [424]: "message not sent"
+    """
+    DB = DataStorage(org_id)
+    message_obj = Message(**request.dict(), org_id=org_id, room_id=room_id,
```
zxenonx, line 52: Here, the server returns an error 500 because the validation failed. Message object needs a `message_id`. 
You can take a look at Room schema in `rooms.py`.
`message_id` is populated only after the message object is stored in the DB, so by default message_id should be None.

backend/endpoints/messages.py
```diff
@@ -1,3 +1,78 @@
...
+                            sender_id= sender_id)
+    response = await DB.write(MESSAGE_COLLECTION, message_obj.dict())
```
zxenonx, line 54: Here, the server returns:
```
{
  "detail": {
    "Message not sent": {
      "status_code": 200,
      "message": {
        "status": 406,
        "message": "maximum collections limit reached"
      }
    }
  }
}
```
Is that what you have gotten ?

backend/requirements.txt
```diff
@@ -55,7 +55,6 @@ trio==0.19.0
 typing-extensions==3.10.0.2
 urllib3==1.26.7
 uvicorn==0.15.0
-uvloop==0.16.0
```
zxenonx, line 58: An issue with the uvloop package? You removed it...

backend/utils/db.py
zxenonx, line 59: More explicit, right?

backend/tests/endpoints/test_message.py
```diff
@@ -2,3 +2,45 @@
...
+        "files": [],
+        "saved_by": [],
+        "is_pinned": False,
+        "is_edited": False,
+        "created_at": "2021-12-11 01:44:37.582879",
+        "threads": []
+        }
+
+def test_send_message():
+    """ test for the send message endpoint"""
```
zxenonx, line 22: First of all, thanks a lot for the tests, @AI-fae. 
The external API call is not that bad. The thing is the API call can fail for many reasons and can be slow...
The purpose here is just to isolate any dependencies and focus on the main logic.
We should then mock these API calls. 
This way we will test the logic and not rely to the API behaviours (just mimicking it)

zxenonx (changes requested):
Good work @AI-fae.
Thanks for the tests again.

backend/endpoints/messages.py
FlorenceEgwu: It appears I got sleepy at some point and salient details flew over my head. 
Thanks for pointing this out.

 I have made the necessary changes.

backend/endpoints/messages.py
FlorenceEgwu: fixed

backend/schema/message.py
FlorenceEgwu: sure!

backend/tests/endpoints/test_message.py
FlorenceEgwu: Alright. 
Will work on the mock function later tonight.

backend/endpoints/messages.py
zxenonx: `if response and response.get("status_code") is None`

backend/endpoints/messages.py
FlorenceEgwu: For some reason I am unaware of, the pre-commit config didn't work for me.
Was wondering if someone took it off or something.

Thanks for the review. 
I have effected the changes.

backend/schema/message.py
FlorenceEgwu: fixed boss

backend/schema/message.py
FlorenceEgwu: Docstring added

backend/endpoints/messages.py
FlorenceEgwu: Well, I could rename it.
But doesn't this negate the structure where the message is a child class of thread?

backend/endpoints/messages.py
FlorenceEgwu: fixed. thank you

backend/endpoints/messages.py
FlorenceEgwu: yes. 
I have reported the issue in the slack channel.
I think an architecture meeting has been scheduled to resolve it.

backend/requirements.txt
FlorenceEgwu: Yes. Couldn't install it on windows.
Thought I could take it off to do my work, then request someone with a compatible OS to add it back

backend/utils/db.py
FlorenceEgwu: yes.
The other response simply said "OK".
Not the kind of error log that could help anybody.

backend/tests/endpoints/test_message.py
FlorenceEgwu: Thank you Zxenon.
I am not familiar with mock testing. I will research and implement it later in the day.

FlorenceEgwu:
Thanks for taking out the time to review my work @zxenonx and @o4codes .
I have effected the requested changes.

Regarding the test, I will have to do a little research on how to make it work. Will appreciate any helpful resources.

backend/endpoints/messages.py
zxenonx: You should not be allowed to write API endpoints after a certain time... @AI-fae

backend/endpoints/messages.py
zxenonx: Oh, ok. Just seen. Thanks. I thought I had activated notifications for that thread...

backend/endpoints/messages.py
zxenonx: @AI-fae, please it's better to use `is` over `==` when comparing to `None`

backend/requirements.txt
zxenonx: Oh, understood. We will find a permanent fix.

zxenonx:
> Thanks for taking out the time to review my work @zxenonx and @o4codes . I have effected the requested changes.
> 
> Regarding the test, I will have to do a little research on how to make it work. Will appreciate any helpful resources.

Thanks @AI-fae. You can take a look at this article [FastAPI testing - Pytest](https://testdriven.io/blog/fastapi-crud/#get-routes)

man-of-mind:
> Thanks for taking out the time to review my work @zxenonx and @o4codes .
> I have effected the requested changes.
> 
> Regarding the test, I will have to do a little research on how to make it work. Will appreciate any helpful resources.
> 

You can check zc_dm_fast on o4codes' repo, there is a test file in the repo, it might be helpful.

backend/tests/endpoints/test_message.py
zxenonx: No problem @AI-fae.
Here are some articles that might help:
- https://testdriven.io/blog/fastapi-crud/#get-routes
- https://www.patricksoftwareblog.com/monkeypatching-with-pytest/

backend/endpoints/messages.py
zxenonx: Hmm, I don't think so... Especially if we are going to use the same endpoint to send a message in a thread.

backend/endpoints/messages.py
o4codes: yea I agree with @zxenonx 
@AI-fae please kindly refactor to this

backend/endpoints/messages.py
o4codes: Well this also works.

backend/endpoints/messages.py
FlorenceEgwu: fixed

backend/endpoints/messages.py
FlorenceEgwu: fixed

o4codes (approved):
Good work

zxenonx:
Thanks for the changes @AI-fae.

zxenonx (commented):
Approved

PR #45, opened 2021-12-14
Minor modifications on messaging

FlorenceEgwu:
- mocked send message test
- changed room and message collection names to fit the recent architecture rules


backend/tests/endpoints/test_message.py
```diff
@@ -1,46 +1,75 @@
...
-        "is_edited": False,
-        "created_at": "2021-12-11 01:44:37.582879",
-        "threads": []
-        }
+fake_org_id = "42893209343920"
+fake_message_obj = {
+    "sender_id": "425324132312",
+    "room_id": fake_room_id,
+    "org_id": fake_org_id,
+    "message_id": "34o003409",
```
man-of-mind, line 15: The message_id should be gotten from zc core and not in the request body

backend/tests/endpoints/test_message.py
```diff
@@ -1,46 +1,75 @@
...
+    monkeypatch.setattr(DB, "write", mock_post)
 
     fake_response_output = {
-        "message_id": "33090940900",
-        "message_data": fake_message_data
+        "room_id": "61b3fb328f945e698c7eb396",
+        "message_id": "61696f43c4133ddga309dcf6",
+        "text": "str",
+        "files": "HTTP url(s)",
+        "sender_id": "619ba4671a5f54782939d385",
```
man-of-mind, line 41: The sender_id is not the same with the one in the request, this can be another test case when the sender_id is not in the room, the class validator in thread function should have flagged it.

man-of-mind (commented):
Well done boss, you could add more test cases to the test. I left some comments as well

FlorenceEgwu:
There's no more need for these changes. As there've been integrated into other PRs

PR #50, opened 2021-12-17
Update message endpoint

paulinebanye:
Created an endpoint to edit a specific message in the collection.
Checks for valid org_id, room_id, sender_id and message_id.

Pending:
Tests


paulinebanye:
@man-of-mind @zxenonx

FlorenceEgwu (changes requested):
Well done @pauline-banye
you've done a really great job here.


Kindly read through some of the reviews below.

backend/schema/message.py
```diff
@@ -81,3 +81,12 @@ class Message(Thread):
...
     threads: List[Thread] = []
+
+
+class MessageUpdateRequest(BaseModel):
+    """
+    Provides a base model for all threads
+    """
+
+    text: str
+    edited_at: str = str(datetime.now())
```
FlorenceEgwu: I think you should use the UTC timezone here. For the purpose of consistency.

str(datetime.utcnow())

backend/endpoints/messages.py
```diff
@@ -74,6 +76,97 @@ async def send_message(
...
         )
     raise HTTPException(
-        status_code=status.HTTP_424_FAILED_DEPENDENCY,
-        detail={"Message not sent": response},
-    )
+            status_code=status.HTTP_424_FAILED_DEPENDENCY,
+            detail={"Message not sent": response},
+        )
+
+
```
FlorenceEgwu, line 80: kindly activate the pre-commit hook by running the startup.sh script.
would help with some of the code formattings.

backend/schema/message.py
```diff
@@ -1,8 +1,7 @@
 import asyncio
 import concurrent.futures
 from datetime import datetime
-from typing import List
-
+from typing import List, Optional
```
FlorenceEgwu: Optional imported but not used

backend/endpoints/messages.py
```diff
@@ -74,6 +76,97 @@ async def send_message(
...
+                status.HTTP_200_OK: {"description": "message updated"},
+                status.HTTP_404_NOT_FOUND: {"description": "message not found"},
+                status.HTTP_403_FORBIDDEN: {"description": "you are not authorized to edit this message"},
+                status.HTTP_424_FAILED_DEPENDENCY: {"description": "message not updated"}
+            })
+async def update_message(
+    request: MessageUpdateRequest,
+    org_id: str,
+    room_id: str,
+    sender_id: str,
```
FlorenceEgwu: please, how is the sender_id passed to the path function?
It's neither in the URL nor request schema.

backend/endpoints/messages.py
```diff
@@ -74,6 +76,97 @@ async def send_message(
...
+                "edited_at": "2021-12-17 11:47:22.678046"
+            }
+            
+    Raises:
+        HTTP_404_FAILED_DEPENDENCY: Message not found
+        HTTP_424_FAILED_DEPENDENCY: Message not edited
+        HTTP_403_FORBIDDEN: You are not authorized to edit this message
+    """
+    DB = DataStorage(org_id)
+    if org_id and room_id and message_id is None:
```
FlorenceEgwu: would this condition ever be fulfilled?
If the aim is to validate that the parameters are valid object ids, might be best to:
1. do this in a different function to keep the endpoint clean. (see something similar in the Room class) 
2. or use the fastapi path validator to achieve this.

FlorenceEgwu (changes requested):

backend/endpoints/messages.py
paulinebanye: Added to schema

backend/endpoints/messages.py
paulinebanye: Noted. I believe this is more of a preference thing tbh. But for consistency, I'll look into making adjustments

pauline-banye: removed

paulinebanye: fixed

backend/schema/message.py
```diff
@@ -81,3 +81,13 @@ class Message(Thread):
     """
 
     threads: List[Thread] = []
+
+
+class MessageUpdate(BaseModel):
```
o4codes, line 86: I feel the MessageRequest class can still be reused though for updating a message. It contains enough necessary fields and also fastapi provides a way of using only fields supplied in the body of requests. 

https://fastapi.tiangolo.com/tutorial/body-updates/

This will enable us to have fewer reusable classes

backend/endpoints/messages.py
```diff
@@ -1,9 +1,10 @@
 from fastapi import APIRouter, BackgroundTasks, HTTPException, status
-from schema.message import Message, MessageRequest
+from schema.message import Message, MessageRequest, MessageUpdate
 from schema.response import ResponseModel
 from starlette.responses import JSONResponse
 from utils.centrifugo import Events, centrifugo_client
 from utils.db import DataStorage
+from utils.mssg_utils import get_mssg
```
o4codes: Please mssg_utils be renamed to message_utils
and also get_mssg to get_message

backend/utils/mssg_utils.py
```diff
@@ -0,0 +1,22 @@
+from utils.db import DataStorage
+
+MESSAGE_COLLECTION = "messages"
+
+
+async def get_mssg(org_id: str, room_id: str, message_id: str) -> dict:
```
o4codes: I feel renaming to get_message is quite intuitive and also concise.

backend/endpoints/messages.py
```diff
@@ -77,3 +78,92 @@ async def send_message(
...
+                content=ResponseModel.success(
+                    data=data, message="message edited"
+                ),
+                status_code=status.HTTP_200_OK,
+            )
+        raise HTTPException(
+            status_code=status.HTTP_424_FAILED_DEPENDENCY,
+            detail={"message not edited": edit_message},
+        )
+    except Exception as e:
```
o4codes: is there any error the try block will get.
Also if an error is caught what response will be returned ?

o4codes (changes requested):
@pauline-banye 
Nice work.

Kindly go through the reviews

backend/schema/message.py
mikenrowland: Also @zxenonx @o4codes I think the message request class should include an 'edited' field set to false by default.
IMO the 'edited at' field can be left out since we are not doing much with that info.
Taking these into consideration the need for an extra class will be eliminated.

backend/endpoints/messages.py
```diff
@@ -77,3 +78,92 @@ async def send_message(
...
+    if not message:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND, detail="Message not found"
+        )
+
+    payload = request.dict()
+    if message["sender_id"] != payload["sender_id"]:
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail="You are not authorized to edit this message",
```
mikenrowland, line 136: I think the appropriate status code for this should be "HTTP_401_UNAUTHORIZED"

backend/endpoints/messages.py
```diff
@@ -77,3 +78,92 @@ async def send_message(
...
+            {
+                "room_id": "619e28c31a5f54782939d59a",
+                "message_id": "61bc5e5378fb01b18fac1426",
+                "sender_id": "619ba4671a5f54782939d385",
+                "text": "testing edits",
+                "edited_at": "2021-12-17 11:47:22.678046"
+            }
+
+    Raises:
+        HTTP_404_FAILED_DEPENDENCY: Message not found
```
mikenrowland: I guess this must be a typo 'HTTP_404_FAILED_DEPENDENCY'

mikenrowland (changes requested):
Well done @pauline-banye 
I left few comments.

backend/endpoints/messages.py
paulinebanye: noted

backend/endpoints/messages.py
paulinebanye: oh yeah it is. Will fix

backend/endpoints/messages.py
pauline-banye: noted

backend/schema/message.py
pauline-banye: I thought about using the MessageRequest class initially but I wasn't too keen on editing someone else's code. However I can simply remove the edited at field, that should work too

paulinebanye: @mykie88 I took out the edited_at field. 
@o4codes unfortunately I did not adjust the endpoint to use the messagerequest class because it would require me editing that class. Re the  method fastapi provides, maybe you could set up a meet and walk me through.

backend/endpoints/messages.py
paulinebanye: Yes it would raise a failed dependency error. I have made it more explicit though

backend/endpoints/messages.py
```diff
@@ -77,3 +76,99 @@ async def send_message(
...
+):
+    """
+    Update a message
+
+    Args:
+        request: Request object
+        org_id: A unique identifier of the organization.
+        room_id: A unique identifier of the room.
+        sender_id: A unique identifier of the sender.
+        message_id: A unique identifier of the message that is being edited.
```
zxenonx: `background_tasks` should be added and `sender_id` should be removed from the Args section

backend/endpoints/messages.py
o4codes: Ohh
Actually No.
It would return the raised http exception (failed dependency) as a response to the endpoint and wouldn't actually cause an program exception to the code itself.

backend/endpoints/messages.py
```diff
@@ -77,3 +76,98 @@ async def send_message(
...
+        raise HTTPException(
+            status_code=status.HTTP_424_FAILED_DEPENDENCY,
+            detail={"message not edited": edit_message},
+        )
+    except HTTPException as e:
+        return JSONResponse(
+            data="Failure to retrieve data",
+            status=status.HTTP_424_FAILED_DEPENDENCY,
+            detail=e
+        )
```
zxenonx: Do we really need a try-except block here? If we do, then an `else` clause is needed because the `try` block should only contains codes that will throw expected errors.
Also the except Exception it a bit too broad. We should catch more concise exceptions.
But as it it now, most of the expected errors are already handled in the code (by raising HTTPException).
The other unexpected errors should fail loudly.

zxenonx (changes requested):
Thanks @pauline-banye for this great work.

backend/endpoints/messages.py
```diff
@@ -77,3 +76,98 @@ async def send_message(
         status_code=status.HTTP_424_FAILED_DEPENDENCY,
         detail={"Message not sent": response},
     )
+
+
+@router.put(
```
zxenonx, line 81: `status_code` is missing inside `@router.put()`.
@pauline-banye

pauline-banye: @zxenonx it's in responses. I didn't see the need to duplicate it.

backend/endpoints/messages.py
paulinebanye: adjusted.

PR #52, opened 2021-12-20
Update message endpoints

paulinebanye:
Created an endpoint to edit a specific message in the collection.
Checks for valid org_id, room_id, and message_id to retrieve message
Ensures that only the sender of the message can edit it.

Pending:
Tests


paulinebanye:
Closed the old one because I felt there were too many commits.
Took everyone's suggestions into consideration and made adjustments. 

@man-of-mind @zxenonx @AI-fae @o4codes @mykie88

o4codes:
@zxenonx
Kindly look through this PR, as I made some contributions to it.

@AI-fae 
Please go through the PR, I changed the endpoint route of the send message to /orgs/:org_id/rooms/:room_id/messages
Therefore sender_id would be passed in the body of the request. This would enable the same class be used for the update_message endpoint.

paulinebanye:
@o4codes Good job on the edits. Both endpoints work without issues.

man-of-mind (approved):
Well done bosses @pauline-banye @o4codes

FlorenceEgwu (approved):
Good work!
 @o4codes and @pauline-banye.

o4codes (approved):

backend/utils/db.py
```diff
@@ -107,14 +107,14 @@ async def read(
...
-            "options": options,
-        }
+        request_body = dict(
+            collection_name=collection_name,
+            filter=query,
+            object_id=resource_id,
+            organization_id=self.organization_id,
+            plugin_id=self.plugin_id,
+            options=options,
+        )
```
zxenonx: It's generally preferable to use {} notation to declare a dict. This is more "pythonic". 
Moreover, from a performance view, declaring a dictionary with dict() is much slower than with {}.

zxenonx (approved):
Thanks @pauline-banye and @o4codes.

backend/utils/db.py
o4codes: Thanks.
@pauline-banye 
Could you help resolve this ?

paulinebanye: Hi @o4codes, I could look at this but not immediately unfortunately.

Currently out n I won't be home till much later tonight and it definitely won't get sorted before the FE meeting

zxenonx: I've made the change. @pauline-banye @o4codes

PR #58, opened 2021-12-23
Adjustments to the message schema and endpoints

paulinebanye:
Message schema was edited to match the request from the FE
Data returned from the update endpoint was switched to the entire edited message object
Minor modifications to the send message endpoint
Updated the docstrings to reflect the new message object
Sample message model was added to schema file for review.


backend/endpoints/messages.py
```diff
@@ -143,18 +194,13 @@ async def update_message(
...
 
     if edited_message and edited_message.get("status_code") is None:
-        new_data = {
-            "room_id": room_id,
-            "message_id": message_id,
-            "sender_id": payload["sender_id"],
-            "text": payload["text"],
-            "edited": True,
-        }
+        new_message = await get_message(org_id, room_id, message_id)
```
man-of-mind: Why do we need to get the message back?

pauline-banye: That's what Aniebet requested yesterday. He said he wanted the entire object to be returned and published to centrifugo.

pauline-banye: @man-of-mind That was actually requested by Aniebet. He said that the entire updated object should be returned and published to centrifugo instead of just the updated payload.

backend/endpoints/messages.py
```diff
@@ -56,24 +77,23 @@ async def send_message(
     response = await DB.write(
         MESSAGE_COLLECTION, message_obj.dict(exclude={"message_id"})
     )
-    message_obj.message_id = response.get("data").get("object_id")
 
     if response and response.get("status_code") is None:
         message_obj.message_id = response["data"]["object_id"]
```
zxenonx, line 82: Thanks @pauline-banye for the change here. 
We should check that response is not None before querying for data value.

backend/endpoints/messages.py
paulinebanye: Sorted boss

backend/schema/message.py
```diff
@@ -1,34 +1,96 @@
...
-    text: str
     sender_id: str
-    reactions: List[Reaction] = []
+    emojis: List[Emoji] = []
+    richUiData: Any = {}
     files: List[AnyHttpUrl] = []
     saved_by: List[str] = []
-    edited: bool = False
+    timestamp: int = int(datetime.utcnow().timestamp())
     created_at: str = str(datetime.utcnow())
```
zxenonx: There's no field to hold the text message. As a result, when we test the endpoint, we can actually send a payload successfully without a text message. 
It should not be possible, right?

zxenonx (commented):
Thanks @pauline-banye. Just left a few comments.

backend/schema/message.py
```diff
@@ -29,7 +29,7 @@ class MessageRequest(BaseModel):
     richUiData: Any = {}
     files: List[AnyHttpUrl] = []
     saved_by: List[str] = []
-    timestamp: int = int(datetime.utcnow().timestamp())
+    timestamp: int
```
zxenonx, line 32: Why didn't you keep the previous default value?

paulinebanye: Hi @zxenonx That was what the @t_a specified. He requested it be left as an integer instead of the default timestamp I set initially. 

It is also a required field so that it throws an error from the frontend if there's any instance that the data is sent without a timestamp

backend/endpoints/messages.py
paulinebanye: @zxenonx Umm, I didn't actually work on this but no worries, I'll adjust it. I guess we missed it during the sprint last night. 

Right now, I'm just waiting for everyone to give their reviews so I don't have too many commits.

backend/schema/message.py
paulinebanye: @zxenonx There's actually field for text in the richUIdata but yes I share your concern about using Any to send anything. 

We need to have a discussion, possibly with Aniebiet, regarding the current structure that's sent as the payload.

zxenonx: yeah, @pauline-banye, the text field is in `richUIdata>blocks`.
So maybe for testing purposes, we could change the actual structure to this one:
```
{
  "sender_id": "619ba4671a5f54782939d385",
  "emojis": [],
  "richUiData": {,
      "blocks": [
                        {
                        "text": "HI, I'm mark.. new here",
                        }
                    ], 
},
  "files": [],
  "saved_by": [],
  "timestamp": 1640261761,
  "created_at": "2021-12-23 13:16:01.273083"
}
```
And `blocks` could be a nested schema inside richUIdata. Too messy?

backend/endpoints/messages.py
zxenonx: I meant you have already fixed that. By moving this line
`message_obj.message_id = response.get("data").get("object_id")` inside the if-statement scope.
@pauline-banye

backend/schema/message.py
zxenonx: Understood @pauline-banye. Thanks.

backend/schema/message.py
paulinebanye: I'ld rather we get on a call for this esp with the team so as to avoid going back n forth on this. Let me know when you're free.

backend/endpoints/messages.py
paulinebanye: Oh okay. I didn't fix it tho. Aniebiet did. Honestly haven't really looked at that endpoint since the sprint.

backend/schema/message.py
zxenonx: @AI-fae @mykie88 @man-of-mind @o4codes @kinglighthill

FlorenceEgwu: I agree, using "any" to receive data could be a weak security point. A couple of things could go wrong.


@zxenonx , it should be possible to send a message without a text. Using WhatsApp and Slack as an example, you can send a file attachment or an emoji character without any text.

paulinebanye: @AI-fae very true re the sending of a file as attachment without any text. 

Yeah I do have concerns about security but based off our conversation in the group, and the reluctance to create non reusable classes to map every single thing, I believe we've decided to stick with Any. 

@zxenonx @man-of-mind @o4codes @mykie88 @kinglighthill

mikenrowland: I think we are all in agreement that mapping everything is not ideal. So "Any" is fine and from the assignment operator, it is by default a dictionary I guess, so we know the kind of data to expect. Besides these, I can't think of any security issue that can ensue from the current set up.
That's my take though...
@zxenonx @pauline-banye

zxenonx: richUIdata is by default a dict, so, `dict` would be fine for the type instead of `Any`.
@pauline-banye

FlorenceEgwu (approved):
Good job @pauline-banye
Compliments of the season

backend/utils/db.py
```diff
@@ -1,216 +1,208 @@
...
-        except requests.exceptions.RequestException as exception:
-            print(exception)
+import requests
+
+PLUGIN_KEY = "chat.zuri.chat"
+BASE_URL = "https://staging.api.zuri.chat"
+
+
+class DataStorage:
+    """Serves as a layer for communication for plugin data and server
```
mykie88, line 8: Hi @pauline-banye please is there any reason why db.py file was touched?

backend/endpoints/messages.py
```diff
@@ -129,8 +178,9 @@ async def update_message(
             status_code=status.HTTP_404_NOT_FOUND, detail="Message not found"
         )
 
-    request.edited = True
     payload = request.dict(exclude_unset=True)
+    message["richUiData"] = payload["richUiData"]
+    payload["edited"] = True
 
     if message["sender_id"] != payload["sender_id"]:
```
mikenrowland, line 182: @pauline-banye sorry please, was doing last minute checks before clicking approved and saw this line. Is this check supposed to come after line 182 or before?

mikenrowland (commented):
@pauline-banye

backend/endpoints/messages.py
paulinebanye: Hi @mykie88,

Lines 182 and 183 simply outline the payload data. 183 changes the "edited" field to True and 182 swaps the original data with the payload data. This was added so that the full updated object was returned in the response, as Aniebiet specified, without having to read zccore again.

mikenrowland: @pauline-banye what I am saying is if message["sender_id"] != payload["sender_id"] on line 185 should come before swapping the "richUiData" on line 182.

man-of-mind: @mykie88 it's not really a major concern as it won't be updated to zc core if the sender_ids are not the same but we can move it to be inside the if block.

paulinebanye: @mykie88 Yeah I understood what you're asking. 

It didn't seem particularly relevant since all it's doing is specifying the payload data, but if it's an issue for you, it can be moved. It doesn't affect the outcome either way

mikenrowland: oh yeah @man-of-mind it is not a major issue except for the little extra memory allocation wasted during the swap if the check eventually fails downstream.
Pardon me if I appear to be too pedantic. The code can be left as is then.
well done @pauline-banye

mikenrowland: Please kindly mark conversation as resolved @pauline-banye

man-of-mind: Yeah, you are right boss @mykie88 , i suggest we move it inside the if block.

mikenrowland (approved):
@zxenonx this LGTM, can we merge it already. FE are expectant, thanks.

backend/endpoints/messages.py
zxenonx: Like @mykie88, I also think that the swap should come after the if check. This may seem particularly irrelevant, but from a logical point of view, it makes sense.
@pauline-banye

paulinebanye: Adjusted @mykie88

zxenonx (approved):
LGTM. Thanks a lot @pauline-banye.

PR #60, opened 2021-12-24
Close conversation endpoint

mikenrowland:
Closes a DM or Group_DM room on the sidebar
By toggling the "closed" boolean field on the room document from False to True
The function when called on a closed room, changes the closed field back to False


backend/endpoints/rooms.py
```diff
@@ -170,3 +170,95 @@ async def join_room(
...
+            status_code=status.HTTP_403_FORBIDDEN, detail="cannot close a channel room"
+        )
+
+    if member_id not in room["room_members"].keys():
+        raise HTTPException(
+            status_code=status.HTTP_401_UNAUTHORIZED, detail="member not in room"
+        )
+
+    member_room_data = room["room_members"].get(member_id)
+    member_room_data["closed"] = member_room_data["closed"] is False
```
man-of-mind, line 236: @mykie88 boss, can u pls explain this line to me?

mikenrowland: okay boss, member_room_data["closed"] is set to "False" by default, so instead of writing too many lines of code just to change it to "True", I decided to just ask if it is False ("member_room_data["closed"] is False") which will automatically return a true and that will be assigned to the "closed" value. 
The same logic will be applied in the reversed order if the closed value of a room is True, so that a previously closed room can be reopened.
Please let me know if it is clear @man-of-mind

zxenonx: We could also write :
```
member_room_data["closed"] = not member_room_data["closed"]
```
@mykie88

backend/endpoints/rooms.py
```diff
@@ -170,3 +170,95 @@ async def join_room(
...
+    data = {"room_members": room["room_members"]}
+
+    update_response = await DB.update(
+        ROOM_COLLECTION, document_id=room_id, data=data
+    )  # updates the room data in the db collection
+
+    background_tasks.add_task(
+        centrifugo_client.publish,
+        room=room_id,
```
FlorenceEgwu: Is everyone in the room to be notified of a closed room?
Thought it's peculiar to the user who closed the room?

backend/endpoints/rooms.py
FlorenceEgwu: I got confused by this line too.
Thanks for the clarification @mykie88 

dope logic!

man-of-mind: > okay boss, member_room_data["closed"] is set to "False" by default, so instead of writing too many lines of code just to change it to "True", I decided to just ask if it is False ("member_room_data["closed"] is False") which will automatically return a true and that will be assigned to the "closed" value. 
> The same logic will be applied in the reversed order if the closed value of a room is True, so that a previously closed room can be reopened.
> Please let me know if it is clear @man-of-mind 

Thanks boss @mykie88

backend/endpoints/rooms.py
mikenrowland: no everyone is not supposed to be notified. does the code suggest that?
the publish is just supposed to do a sidebar update so that the sidebar details for that particular member will be updated so that the room can be excluded.
Please tell me if it seems different @AI-fae

mykie88: the notification is just for that particular member, does the  code suggest otherwise? please let me know. 
the sidebar update publish is for updating the sidebar details of that particular member so that the closed room is removed from the joined rooms list.

backend/endpoints/rooms.py
```diff
@@ -170,3 +170,95 @@ async def join_room(
...
+                            "closed": true || false,
+                            "role": "admin",
+                            "starred": false
+                        }
+                    }
+    Raises:
+        HTTP_401_UNAUTHORIZED: member not in room
+        HTTP_403_FORBIDDEN: cannot close a channel room
+        HTTP_404_NOT_FOUND: room not found
+        HTTP_424_FAILED_DEPENDENCY: failed to add new members to room
```
paulinebanye: Line 215. Change to unable to close conversation

mikenrowland: thanks @pauline-banye

backend/endpoints/rooms.py
FlorenceEgwu: publishing to the room id will publish to every member of the room if I remember correctly.

If you want to update just the user's sidebar, you might want to publish to the user id instead. Ref the create_room endpoint to see how this was done.

FlorenceEgwu: In the sidebar.py file, there's a helper function for updating the sidebar on centrifugo. You should take a look at it too...``

FlorenceEgwu (changes requested):

backend/endpoints/rooms.py
mikenrowland: I'm not sure an event can be published to a user_id. 
But I will take a look at the suggested files and see what can be done. Thanks

backend/endpoints/rooms.py
```diff
@@ -170,3 +170,95 @@ async def join_room(
...
+    )  # updates the room data in the db collection
+
+    background_tasks.add_task(
+        centrifugo_client.publish,
+        room=room_id,
+        event=Events.SIDEBAR_UPDATE,
+        data=member_room_data,
+    )  # publish to centrifugo in the background
+
+    if update_response and update_response.get("status_code", None) is None:
```
zxenonx:
```suggestion
    if update_response and update_response.get("status_code") is None:
```
We can remove the second argument of get(). By default it returns None.

backend/utils/sidebar.py
```diff
@@ -117,10 +117,11 @@ async def __get_joined_rooms(
         starred_rooms = []
         user_rooms = user_rooms or []
         for room in user_rooms:
-            room_profile = await self.__get_room_profile(
-                member_id, room, org_members, DB
-            )
-            rooms.append(room_profile)
+            if room.get("room_members").get(member_id).get("closed") is False:
```
zxenonx, line 120: We could also write it this way:
```suggestion
            if not room.get("room_members").get(member_id).get("closed"):
```

backend/utils/sidebar.py
```diff
@@ -117,10 +117,11 @@ async def __get_joined_rooms(
...
+                room_profile = await self.__get_room_profile(
+                    member_id, room, org_members, DB
+                )
+                rooms.append(room_profile)
             if room.get("room_members").get(member_id, {}).get("starred", None):
```
zxenonx: As suggested above: 
```suggestion
            if room.get("room_members").get(member_id, {}).get("starred"):
```

zxenonx (commented):
Thanks @mykie88 for the work.

backend/utils/sidebar.py
mikenrowland: is there a reason why using "not" is better or is it just a matter of preference?

backend/endpoints/rooms.py
mikenrowland: True, will fix that

backend/utils/sidebar.py
mikenrowland: seen

backend/utils/sidebar.py
zxenonx: The pythonic way is to not explicitly compare to True /False if we don't need to, as it's the case here.

Also, negating an expr seems faster than comparing it to False. 

Kindly take a look at this article :

https://switowski.com/blog/checking-for-true-or-false

@mykie88

mikenrowland: thanks alot @zxenonx 
I've read the article and it is in agreement with the current code structure because we are explicitly checking for the True or False value of "closed". 
Here is the referenced recommendation from the article:
"**If you want to check that a variable is explicitly True or False (and is not truthy/falsy), use is (if variable is True)**"

zxenonx: Yeah @mykie88, I can see you read it fully.
And obviously, you're not wrong at all. 
I just find the "if not" expr more straightforward.
But what you've written is absolutely right, so there's no problem at all.

zxenonx (approved):
Thanks @mykie88 for the changes.
LGTM. 
Can we merge this?

FlorenceEgwu (approved):
@mykie88 thanks for making the requested changes.

well done!

backend/utils/sidebar.py
mikenrowland: Thanks boss 🙏

PR #61, opened 2021-12-28
Feature/get room messages endpoint

zxenonx:
- Implement an API endpoint to fetch all messages inside a particular room
- Refactor utility function get_room_messages()


backend/endpoints/messages.py
```diff
@@ -204,3 +205,27 @@ async def update_message(
...
+
+    Returns:
+        A list of message objects
+
+    """
+    room_messages = await get_room_messages(org_id, room_id)
+    return JSONResponse(
+        content=ResponseModel.success(data=room_messages, message="Messages retrieved"),
+        status_code=status.HTTP_200_OK,
+    )
```
zxenonx: So, there's a utility function (get_room_messages) that helps to retrieve messages in a room as a list via DataStorage.
Here's the code:

```
DB = DataStorage(org_id)
response = await DB.read(MESSAGE_COLLECTION, query={"room_id": room_id})
    if response and "status_code" not in response:
        return response
    return {}

```
The issue is that, inside the endpoint, there's no way I can raise an exception (for example when Zc core fails). 
In fact, in the code above, the function returns either a list of messages or an empty dict.
When I test the get messages endpoint with a non-valid room, or a room in which there are no messages yet, I always get an empty dict. I guess I'll get the same result when Zc core will fail. If there are no messages, we should get an empty array, not an empty dict.
This is really ambiguous.

The main idea I have in mind is to refactor the utility function so that it will only return the raw response without any checking. Then, the necessary checks will be done in the endpoint function.

What do you think about this?

@mykie88 @pauline-banye @AI-fae @man-of-mind @kinglighthill

zxenonx: @o4codes

paulinebanye: I think you can go ahead with the refactor.

man-of-mind: You can go ahead with the refactor @zxenonx , it's better that way.

zxenonx: Ok. Thanks. I'll do that then.

man-of-mind (approved):
Well done boss

backend/endpoints/messages.py
mikenrowland: I think this problem will not be peculiar to this endpoint going forward, because most of those utility functions are returning empty dicts in all cases of a failed response or "falsy" return values.
IMO all possible response types (success, failed, falsy) should be accounted for in all the utility functions.

backend/utils/message_utils.py
```diff
@@ -31,9 +31,8 @@ async def get_room_messages(org_id: str, room_id: str) -> list:
 
     DB = DataStorage(org_id)
     response = await DB.read(MESSAGE_COLLECTION, query={"room_id": room_id})
-    if response and "status_code" not in response:
-        return response
-    return {}
+
+    return response or []
```
o4codes, line 35: Nice one boss

o4codes (approved):
Good work

backend/utils/message_utils.py
```diff
@@ -31,9 +31,8 @@ async def get_room_messages(org_id: str, room_id: str) -> list:
 
     DB = DataStorage(org_id)
     response = await DB.read(MESSAGE_COLLECTION, query={"room_id": room_id})
```
paulinebanye, line 33: Could you add an option for sorting the returned messages?

backend/endpoints/messages.py
```diff
@@ -204,3 +205,32 @@ async def update_message(
...
+
+    Returns:
+        A list of message objects
+
+    """
+    response = await get_room_messages(org_id, room_id)
+    if response is None or "status_code" in response:
+        raise HTTPException(
+            status_code=status.HTTP_424_FAILED_DEPENDENCY,
+            detail=response,
```
paulinebanye: I think the ZC core failed message should be added to the returned response.

backend/utils/message_utils.py
zxenonx: @pauline-banye. Yeah, let's see first what the FE devs would like to add.

backend/endpoints/messages.py
zxenonx: Ok. We'll handle the custom exception later. @pauline-banye

PR #68, opened 2022-01-06
Get Room members

paulinebanye:
Endpoint to retrieve a list of all the members in a room. 
Returns all members in a room if the room is found in the database
Raises HTTP_404_NOT_FOUND if the room is not found
Raises HTTP_424_FAILED_DEPENDENCY if there is an error retrieving the room members


o4codes (changes requested):
What if you added as an extra functionality, ability to add query parameters such as closed, role or  starred.

Therefore, if any of the query paramters is supplied, it will produce room_members using that paramter as a filter.

@pauline-banye NIce work though

backend/endpoints/rooms.py
```diff
@@ -262,3 +262,56 @@ async def close_conversation(
...
+    "/org/{org_id}/rooms/{room_id}/members",
+    response_model=ResponseModel,
+    status_code=status.HTTP_200_OK,
+    responses={
+        424: {"detail": "Failure to retrieve room members"},
+    },
+)
+async def get_members(org_id: str, room_id: str):
+    """
+    Get room members.
```
zxenonx:
```suggestion
    """Get room members.
```

backend/endpoints/rooms.py
```diff
@@ -262,3 +262,56 @@ async def close_conversation(
...
+    Raises HTTP_424_FAILED_DEPENDENCY if there is an error retrieving the room members
+    Args:
+        org_id (str): A unique identifier of an organisation
+        room_id (str): A unique identifier of the room
+    Returns:
+        HTTP_200_OK (Room members)
+
+        {
+            "status": "success",
+            "message": "Room Members",
```
zxenonx: I suggest this: 
```suggestion
            "message": "Room members retrieved successfully",
```

backend/endpoints/rooms.py
```diff
@@ -262,3 +262,56 @@ async def close_conversation(
...
+    if members and members.get("status_code") is None:
+        return JSONResponse(
+            content=ResponseModel.success(data=members, message="Room Members"),
+            status_code=status.HTTP_200_OK,
+        )
+        
+    raise HTTPException(
+        status_code=status.HTTP_424_FAILED_DEPENDENCY,
+        detail="Failed to retrieve room members",
+    )
```
zxenonx: This exception is also raised when an invalid room is provided. 

We should consider this special case and return a "room doesn't exist" message instead.

backend/endpoints/rooms.py
```diff
@@ -262,3 +262,56 @@ async def close_conversation(
...
+
+@router.get(
+    "/org/{org_id}/rooms/{room_id}/members",
+    response_model=ResponseModel,
+    status_code=status.HTTP_200_OK,
+    responses={
+        424: {"detail": "Failure to retrieve room members"},
+    },
+)
+async def get_members(org_id: str, room_id: str):
```
zxenonx, line 275: Kindly add the return type

zxenonx (changes requested):

zxenonx:
Thanks for the work @pauline-banye. 
I've left a few comments.

backend/endpoints/rooms.py
paulinebanye: I switched it to get room instead so room doesn't exist is accounted for as well as accounting for the failed dependency exception.

backend/endpoints/rooms.py
```diff
@@ -353,3 +353,92 @@ async def close_conversation(
...
+        )
+
+    members = room["room_members"]
+    if members and members.get("status_code") is not None:
+        raise HTTPException(
+            status_code=status.HTTP_424_FAILED_DEPENDENCY,
+            detail="Failure to retrieve room members",
+        )
+
+    if keyword:  # if keyword is supplied
```
man-of-mind: Well done boss @pauline-banye 

So the filter option is to only return members or admins in a room, i thought @o4codes was referring to filtering based starred, close and roles, maybe u can include these as well boss.

paulinebanye: Repent o @man-of-mind! 

Re filtering for starred and closed rooms, I had a conversation with both @o4codes and @zxenonx and it was concluded that there's no use case for it. 

It should be done in the room, if at all. They are both Boolean fields without any specific information about rooms being closed or starred. That's why I ended up leaving them out.

man-of-mind: Oh...okay boss

paulinebanye: Do you have a different opinion? Is there a use case you can think of?

man-of-mind: Not really o.....my ogas don talk

backend/endpoints/rooms.py
```diff
@@ -353,3 +353,92 @@ async def close_conversation(
...
+    "/org/{org_id}/rooms/{room_id}/members",
+    response_model=ResponseModel,
+    status_code=status.HTTP_200_OK,
+    responses={
+        400: {"detail": "Invalid keyword supplied"},
+        404: {"detail": "Room not found"},
+        424: {"detail": "Failure to retrieve room members"},
+    },
+)
+async def get_members(org_id: str, room_id: str, keyword: Optional[str] = None):
```
man-of-mind: Can this keyword be role? 
@pauline-banye

backend/endpoints/rooms.py
paulinebanye: If you do have an opinion or a change you want me to implement, pls speak up.

man-of-mind: 😂😂😂😂

No boss, I am serious

o4codes: I feel when an endpoint is created, the possibilities of thaet endpoint should be maxed out.
The concepts of use cases helps in aligning priority.
I feel those extra features could still be added.
But then for us to move faster, it can be waved off, till someone free enough decides to add them
@pauline-banye and @man-of-mind

backend/endpoints/rooms.py
o4codes: In place of keyword, this can be called "role"

paulinebanye: Yes it can, I deliberately didn't use role so its not mixed up with the member["role"] but if it's an ish, I can adjust it.

backend/endpoints/rooms.py
man-of-mind: > I feel when an endpoint is created, the possibilities of thaet endpoint should be maxed out.
> The concepts of use cases helps in aligning priority.
> I feel those extra features could still be added.
> But then for us to move faster, it can be waved off, till someone free enough decides to add them
> @pauline-banye and @man-of-mind 

Well said boss
All extra features can be put on hold for the moment.

backend/endpoints/rooms.py
```diff
@@ -353,3 +353,92 @@ async def close_conversation(
...
+        raise HTTPException(
+            status_code=status.HTTP_424_FAILED_DEPENDENCY,
+            detail="Failure to retrieve room members",
+        )
+
+    if keyword:  # if keyword is supplied
+        if keyword not in ["admin", "member"]:
+            raise HTTPException(
+                status_code=status.HTTP_400_BAD_REQUEST,
+                detail="Invalid keyword supplied",
```
o4codes: So the detail could be "invalid role"

o4codes (commented):
Good work @pauline-banye 
I left some small reviews

mikenrowland (commented):
Well done @pauline-banye

backend/endpoints/rooms.py
paulinebanye: I created the code with the extra filters initially and I still have it in my codebase. 

So I've decided that since the consensus is that it might not be necessary but we would add it later, I'll just push that instead. 

Since majority have an issue with the name of the filter when it checked for just the role, with the new code that has the check for the extra "keywords", what do you suggest I rename it to? 

@zxenonx @o4codes @man-of-mind

backend/endpoints/rooms.py
zxenonx: I don't think we need to filter based on these parameters (closed, starred) for now.

o4codes (approved):
Good work @pauline-banye

man-of-mind (approved):
Well done boss

zxenonx (approved):
Thanks for the work @pauline-banye

PR #71, opened 2022-01-07
Get Channels Endpoint

FlorenceEgwu:
Created an endpoint to fetch all available public channels and joined private channels (if member_id is provided)


FlorenceEgwu:
I will be closing this PR because a close look has revealed that this functionality is already covered in the sidebar endpoint.
PR for get_all rooms coming in a bit

PR #75, opened 2022-01-10
Test for update message endpoint

man-of-mind:
Created test for update message endpoint and did some refactoring.

Kindly review @o4codes @zxenonx @mykie88 @pauline-banye @kinglighthill @AI-fae


backend/tests/endpoints/test_message.py
```diff
@@ -158,8 +205,109 @@ async def test_send_message_check_status_code(mock_get_user_room, mock_write):
...
+    mock_get_user_room.return_value = fake_core_room_data
     mock_write.return_value = write_response
-    response = client.post(send_message_test_url, json=test_payload)
+    response = client.post(send_message_test_url, json=send_message_test_payload)
     assert response.status_code == 424
     assert response.json() == {"detail": {"Message not sent": write_response}}
+
+
+@pytest.mark.asyncio
+async def test_update_message_sucessful(
```
zxenonx:
```suggestion
async def test_update_message_successful(
```

mikenrowland (approved):
LGTM

mikenrowland (approved):

backend/tests/endpoints/test_message.py
```diff
@@ -158,8 +205,109 @@ async def test_send_message_check_status_code(mock_get_user_room, mock_write):
...
+    mock_get_message.return_value = fake_zc_core_message_data
+    mock_update_message.return_value = {
+        "status": 200,
+        "message": "success",
+        "data": {"matched_documents": 1, "modified_documents": 1},
+    }
+    mock_centrifugo.return_value = {"status_code": 200}
+    response = client.put(update_message_test_url, json=update_message_test_payload)
+    assert response.status_code == 200
+    assert response.json() == {
```
zxenonx, line 259: This test fails because of `edited` value in `response.json()`. It should be True as the message has been successfullt edited, right?

zxenonx (changes requested):
Thanks @man-of-mind.

backend/tests/endpoints/test_message.py
man-of-mind: Yes boss @zxenonx but it is returning False in the endpoint.

backend/tests/endpoints/test_message.py
man-of-mind: Okay boss

backend/tests/endpoints/test_message.py
zxenonx: When we'll merge the Pagination PR, we will then fix this.

man-of-mind: Okay boss

backend/tests/endpoints/conftest.py
```diff
@@ -0,0 +1,81 @@
+from unittest.mock import AsyncMock
+
+import pytest
+
+
```
zxenonx, line 5: Great work here. Very helpful.

backend/tests/endpoints/test_message.py
```diff
@@ -92,74 +150,211 @@ async def test_send_message_successful(mock_get_user_room, mock_write, mock_cent
...
+
+
+@pytest.mark.asyncio
+async def test_connection_timeout():
+    """Test for connection timeout"""
+    with mock.patch.object(
+        DataStorage, "__init__", Mock(return_value=exceptions.RequestException)
+    ) as mocker:
+        DataStorage.__init__("12334")
+        mocker.assert_called_with("12334")
```
zxenonx: Can we move this to test_db.py?

backend/tests/endpoints/test_message.py
```diff
@@ -92,74 +150,211 @@ async def test_send_message_successful(mock_get_user_room, mock_write, mock_cent
...
+            "saved_by": [],
+            "sender_id": "619ba4",
+            "threads": [],
+            "timestamp": 0,
+        },
+    }
+
+
+@pytest.mark.asyncio
+@mock.patch.object(DataStorage, "__init__", lambda x, y: None)
```
zxenonx, line 293: New things to learn here. 
Are we mocking only the __init__ method or the whole DataStorage class?

zxenonx (commented):
Great work @man-of-mind. 
I've left a few comments. 
Thanks

backend/tests/endpoints/test_message.py
man-of-mind: This is only mocking the init method in the DataStorage class boss

backend/tests/endpoints/test_message.py
man-of-mind: This can be moved there boss, then we will have to write test for the whole datastorage class, I am thinking we have indirectly written the test for the methods of the DataStorage inside the tests for each endpoints. If need be for us to write the test, no issue then, same way as above, instead of mocking only the init method, we mock the object and have access to all of the methods. @zxenonx

zxenonx: Absolutely agree with you. 
We can mock the whole DataStorage then.

man-of-mind: Okay boss

man-of-mind: Okay boss

zxenonx (approved):
Thanks for the changes @man-of-mind.

LGTM

PR #76, opened 2022-01-12
Join room fix

mikenrowland:
updated the join room endpoint to allow members join public rooms without admin restriction
@zxenonx @AI-fae @man-of-mind @o4codes


paulinebanye (approved):
well done @mykie88

backend/endpoints/rooms.py
```diff
@@ -219,17 +220,19 @@ async def join_room(
...
     member = room.get("room_members").get(str(member_id))
-    if member is None or member["role"].lower() != Role.ADMIN:
-        raise HTTPException(
-            status_code=status.HTTP_401_UNAUTHORIZED,
-            detail="member not in room or not an admin",
-        )
 
     if room["room_type"].upper() == RoomType.CHANNEL:
+        if room["is_private"] is True and (
+            member is None or member["role"].lower() != Role.ADMIN
```
zxenonx, line 231: 👍

backend/endpoints/rooms.py
```diff
@@ -195,17 +195,18 @@ async def join_room(
...
     Returns:
         HTTP_200_OK: {
-                        "status": 200,
-                        "message": "success",
-                        "data": {
-                            "matched_documents": 1,
-                            "modified_documents": 1
-                        }
-                    }
+                "status": 200,
```
zxenonx, line 199: 👍

backend/endpoints/rooms.py
```diff
@@ -195,17 +195,18 @@ async def join_room(
```
zxenonx, line 196: KIndly add `response_model` in @router.put()

zxenonx: Also, in the code below, let's use the ResponseModel class for `content`
```
return JSONResponse(
            status_code=status.HTTP_200_OK,
            content=update_members,
        )
```

backend/endpoints/rooms.py
```diff
@@ -219,17 +220,19 @@ async def join_room(
...
+        if room["is_private"] is True and (
+            member is None or member["role"].lower() != Role.ADMIN
+        ):
+            raise HTTPException(
+                status_code=status.HTTP_401_UNAUTHORIZED,
+                detail="only admins can add new members",
+            )
         room["room_members"].update(members)
 
     if room["room_type"].upper() == RoomType.GROUP_DM:
```
zxenonx, line 239: I've tested the endpoint with these values:
# Parameters
```
org_id == 619ba4671a5f54782939d384
room_id == 61ca29c478fb01b18fac1477 (public channel, with room_members == None)
member_id == 619e28c31a5f54782939d59a (not a member in the room_id's members)
```

# Request body
```
{
  "additionalProp1": {
    "role": "member",
    "starred": false,
    "closed": false
  },
  "additionalProp2": {
    "role": "member",
    "starred": false,
    "closed": false
  },
  "additionalProp3": {
    "role": "member",
    "starred": false,
    "closed": false
  }
}
```

But the endpoint ran successfully with these values.
We should add some additional checks.
What do you think about it? I also suggest another PR to fix this issue.

zxenonx (commented):
Thanks @mykie88 for the work.
I've left a few comments.

backend/endpoints/rooms.py
mikenrowland: wow @zxenonx that's so true o!

mikenrowland: an additional check will complicate this endpoint further sha, because the only way I can think of right now is checking if the member is the organization, which will require fetching all members in the org via a call to zc_core, smh.
do you have a different suggestion? @zxenonx

mikenrowland: but that check is very necessary

zxenonx: Yeah @mykie88 , the check can be done in a separate function and called in the endpoint. 
But for now, I don't see any other solutions than what you suggested.
We can put our heads together and come up with the best way to solve this.

mikenrowland: Yeah that will be great. Thanks @zxenonx

zxenonx: @o4codes @pauline-banye @man-of-mind @kinglighthill @AI-fae

zxenonx (approved):
Thanks for the changes @mykie88.
LGTM.

PR #80, opened 2022-01-13
Tests for Join Room Endpoint

mikenrowland:
Includes test for successfully joining a room, admin successfully adding a member to a private room, restriction from joining a DM room, and test for maximum number of members in a group DM
@zxenonx @man-of-mind @o4codes @AI-fae @pauline-banye


backend/tests/endpoints/test_room.py
```diff
@@ -1,4 +1,127 @@
+import pytest
 from fastapi.testclient import TestClient
 from main import app
 
 client = TestClient(app)
+
+join_room_test_url = "api/v1/org/619org/rooms/619Chrm1/members/619mem1"
+test_payload = {"619mem3": {"role": "member", "starred": False, "closed": False}}
```
paulinebanye:
```suggestion
test_join_room_payload = {"619mem3": {"role": "member", "starred": False, "closed": False}}
```

backend/tests/endpoints/test_room.py
```diff
@@ -1,4 +1,127 @@
...
+fake_core_room_data = {
```
paulinebanye: I think this name is too generic. Consider explicitly renaming this so there's no potential conflict with other peoples tests. 

Except you're creating a fake_core_room_data that is reuseable for everyone's test requirements

backend/tests/endpoints/test_room.py
```diff
@@ -1,4 +1,127 @@
...
+        "data": {"matched_documents": 1, "modified_documents": 1},
+    }
+
+    centrifugo_response = {"status_code": 200}
+
+    mock_get_user_room.return_value = fake_core_room_data
+    mock_update.return_value = update_response
+    mock_centrifugo.return_value = centrifugo_response
+
+    response = client.put(url=join_room_test_url, json=test_payload)
```
paulinebanye:
```suggestion
    response = client.put(url=join_room_test_url, json=test_join_room_payload)
```

backend/tests/endpoints/test_room.py
```diff
@@ -1,4 +1,127 @@
...
```
paulinebanye:
```suggestion
    response = client.put(url=join_room_test_url, json=test_join_room_payload)
```

backend/tests/endpoints/test_room.py
```diff
@@ -1,4 +1,127 @@
...
+async def test_cannot_join_DMroom(mock_get_user_room):
+    """Tests when a member is successfully stopped from joining a DM room
+
+    Args:
+        mock_get_user_room (AsyncMock): Asynchronous external api call
+    """
+    fake_core_room_data["room_type"] = "DM"
+    mock_get_user_room.return_value = fake_core_room_data
+
+    response = client.put(url=join_room_test_url, json=test_payload)
```
paulinebanye:
```suggestion
    response = client.put(url=join_room_test_url, json=test_join_room_payload)
```

backend/tests/endpoints/test_room.py
```diff
@@ -1,4 +1,127 @@
...
+        "619mem5": {"role": "member", "starred": False, "closed": False},
+        "619mem6": {"role": "member", "starred": False, "closed": False},
+        "619mem7": {"role": "member", "starred": False, "closed": False},
+        "619mem8": {"role": "member", "starred": False, "closed": False},
+        "619mem9": {"role": "member", "starred": False, "closed": False},
+        "619mem10": {"role": "member", "starred": False, "closed": False},
+        "619mem11": {"role": "member", "starred": False, "closed": False},
+    }
+    test_payload.update(payload)
+    response = client.put(url=join_room_test_url, json=test_payload)
```
paulinebanye:
```suggestion
    response = client.put(url=join_room_test_url, json=test_join_room_payload)
```

backend/tests/endpoints/test_room.py
```diff
@@ -1,4 +1,127 @@
...
+        "619mem4": {"role": "member", "starred": False, "closed": False},
+        "619mem5": {"role": "member", "starred": False, "closed": False},
+        "619mem6": {"role": "member", "starred": False, "closed": False},
+        "619mem7": {"role": "member", "starred": False, "closed": False},
+        "619mem8": {"role": "member", "starred": False, "closed": False},
+        "619mem9": {"role": "member", "starred": False, "closed": False},
+        "619mem10": {"role": "member", "starred": False, "closed": False},
+        "619mem11": {"role": "member", "starred": False, "closed": False},
+    }
+    test_payload.update(payload)
```
paulinebanye:
```suggestion
    test_join_room_payload.update(payload)
```

paulinebanye (changes requested):

backend/tests/endpoints/test_room.py
mikenrowland: depends on what the person is testing sha, I just used the minimal K-V pairs required for my test cases. But definitely it can be made to replicate the core room object by adding all the possible fields so that other tests can use it, only making minor modifications to the fields in their tests where necessary depending on what they want to test.
As for the name, the idea is for it to be generic, I will update the fields to reflect that.

paulinebanye: I agree on reusing the room data then people can add more information as required but I was considering tests for Channels, Group DMs and DMs. They each have slightly different data. 

For example if this room data is for DM, can it be used for a channel as well? I don't think it can. I think you should specify the room name explicitly. Maybe consider naming it something like fake_dm_room_data

mikenrowland: @pauline-banye nah, I don't think that is necessary, that will mean creating several fake room data for every test case, that will make the entire file messy. If you check the tests cases I covered you will find that I tested for channels, dm and group_dm without having to create different fake room data for each of them. 

So it is not an issue and like I said earlier, the fake room data fields can be modified in each test case to suit what the person wants to test for.

man-of-mind:
@mykie88 please pull and see the changes made in conftest,py file and the additional patch for dataStorage init method in test_messages.py

backend/conftest.py
```diff
@@ -0,0 +1,66 @@
+from unittest.mock import AsyncMock
```
man-of-mind: This file has been moved inside the tests folder boss, there are fixtures for read and update you can use @mykie88

man-of-mind (changes requested):

backend/conftest.py
mikenrowland: omo the merge deleted it o, how come it is here again smh!

mikenrowland: thanks baba @man-of-mind for identifying this, all fixed now. Gracias!!!

man-of-mind (approved):
Well done boss @mykie88

zxenonx (commented):
Great work @mykie88.
There are some edge cases that need to be considered as well.
We actually need to have tests for every condition in if-statements and exceptions.

Tests for:
- Joining a room not found
- Joining a private room as a non-admin member
- Failing to add new members

mykie88:
>

mikenrowland:
@zxenonx I thought of this too but omo if we write tests for all the conditionals per endpoint, the lines of code in the test_room file alone will be over 1000 lines o. We have over 5 room endpoints already and each is packed full with exceptions; from create room to join room, remove from room, close conversation, get room members, etc.

zxenonx:
> @zxenonx I thought of this too but omo if we write tests for all the conditionals per endpoint, the lines of code in the test_room file alone will be over 1000 lines o.

This the essence of writing tests @mykie88. 

Don't worry , 😅, in the end we'll have hundreds of tests and lots of LOC in the test files.

mikenrowland:
alrighty if you say so, I will write them all and update the PR accordingly. Thanks @zxenonx 
btw what is LOC? lol

zxenonx:
> alrighty if you say so, I will write them all and update the PR accordingly. Thanks @zxenonx 
> btw what is LOC? lol

Lines of code, @mykie88

paulinebanye (approved):
well done @mykie88

backend/tests/endpoints/test_room.py
```diff
@@ -1,4 +1,286 @@
...
+    "description": None,
+    "topic": "General Information",
+    "is_private": False,
+    "is_archived": False,
+    "id": "23dg67l0eba8adb50ca13a24",
+    "org_id": "3467sd4671a5f5478df56u911",
+    "created_by": "619ba4671a5f54782939d385",
+}
+
+success_response = {
```
paulinebanye: just noticed this is named success_response. potential conflicts with other codes.

zxenonx: @mykie88

mikenrowland: I don't understand how @pauline-banye

paulinebanye: @mykie88  This success response is particular to only the join room test. It's not re-useable for other tests the way the fake_room_data is and it's not explicitly named.

mikenrowland: oh I get now

zxenonx (approved):
Thanks for the changes.

LGTM.

zxenonx:
Thanks @mykie88.

PR #81, opened 2022-01-14
get room members tests

paulinebanye:
Tests for the get room members endpoint.

@o4codes @zxenonx @man-of-mind @mykie88 @kinglighthill @AI-fae

### DIFF
--- a/backend/conftest.py
+++ b/backend/conftest.py
@@ -49,3 +49,16 @@ def fixture_mock_centrifugo(mocker):
         side_effect=async_mock_centrifugo,
     )
     return async_mock_centrifugo
+
+
+@pytest.fixture(name="mock_dataStorage_read")
+def fixture_mock_dataStorage_read(mocker):
+    """Patch for reading data from zc core.
+    Args:
+       mocker (Mock): For patching a third-party api call
+    Returns:
+       AsyncMock: An instance of the asyncmock class
+    """
+    core_read_mock = AsyncMock()
+    mocker.patch("utils.db.DataStorage.read", side_effect=core_read_mock)
+    return core_read_mock
--- a/backend/tests/endpoints/test_room.py
+++ b/backend/tests/endpoints/test_room.py
@@ -1,4 +1,92 @@
+import pytest
 from fastapi.testclient import TestClient
 from main import app
 
 client = TestClient(app)
+
+
+get_room_members_url = "api/v1/org/619ba46/rooms/61dcf85/members"
+fake_data_group_dm = {
+    "room_name": "test group dm",
+    "room_type": "GROUP_DM",
+    "room_members": {
+        "61696f5": {"role": "admin", "starred": False, "closed": False},
+        "6169704": {"role": "admin", "starred": False, "closed": False},
+        "619ba46": {"role": "admin", "starred": False, "closed": False},
+        "619baa5": {"role": "member", "starred": False, "closed": False},
+    },
+    "created_at": "2022-01-11 03:18:02.364291",
+    "description": None,
+    "topic": None,
+    "is_private": True,
+    "is_archived": False,
+    "id": "61dcf85",
+    "org_id": "619ba46",
+    "created_by": "619ba46",
+}
+
+
+@pytest.mark.asyncio
+async def test_get_room_members_successful(mock_get_user_room):
+    """Tests when room members are retrieved successfully
+    Args:
+        mock_get_user_room (AsyncMock): Asynchronous external api call
+    """
+    members = {
+        "61696f5": {"role": "admin", "starred": False, "closed": False},
+        "6169704": {"role": "admin", "starred": False, "closed": False},
+        "619ba46": {"role": "admin", "starred": False, "closed": False},
+        "619baa5": {"role": "member", "starred": False, "closed": False},
+    }
+    read_response = {
+        "status": "success",
+        "message": "Room members retrieved successfully",
+        "data": members,
+    }
+
+    mock_get_user_room.return_value = fake_data_group_dm
+    response = client.get(url=get_room_members_url)
+    assert response.status_code == 200
+    assert response.json() == read_response
+
+
+@pytest.mark.asyncio
+async def test_get_room_members_room_not_found(mock_get_user_room):
+    """Tests when room is not found
+    Args:
+        mock_get_user_room (AsyncMock): Asynchronous external api call
+    """
+    mock_get_user_room.return_value = None
+    read_response = {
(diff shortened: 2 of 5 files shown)


backend/tests/endpoints/test_room.py
```diff
@@ -1,4 +1,92 @@
...
+@pytest.mark.asyncio
+async def test_get_room_members_status_code(mock_get_user_room):
+    """Tests when room is not found
+    Args:
+        mock_get_user_room (AsyncMock): Asynchronous external api call
+    """
+    mock_get_user_room.return_value = fake_data_group_dm
+
+    response = client.get(url=get_room_members_url)
+    assert response.status_code == 200
```
man-of-mind: Status code is still 200?

backend/tests/endpoints/test_room.py
```diff
@@ -1,4 +1,92 @@
...
+    assert response.json() == read_response
+
+
+@pytest.mark.asyncio
+async def test_get_room_members_status_code(mock_get_user_room):
+    """Tests when room is not found
+    Args:
+        mock_get_user_room (AsyncMock): Asynchronous external api call
+    """
+    mock_get_user_room.return_value = fake_data_group_dm
```
man-of-mind: Not returning status_code in the get user room response?

backend/tests/endpoints/test_room.py
```diff
@@ -1,4 +1,92 @@
...
+    """Tests when room is not found
+    Args:
+        mock_get_user_room (AsyncMock): Asynchronous external api call
+    """
+    fake_url = "api/v1/org/939d384/rooms/0chfy68/members"
+    mock_get_user_room.return_value = fake_data_group_dm
+
+    response = client.get(url=get_room_members_url)
+    assert get_room_members_url != fake_url
+    assert response.status_code == 200
```
man-of-mind: Invalid url will still return 200?

man-of-mind (commented):
Boss @pauline-banye

backend/tests/endpoints/test_room.py
paulinebanye: its asserting that the fake url **isn't** the create room url.

backend/tests/endpoints/test_room.py
pauline-banye: Ummm I don't understand the question. The success response is 200 for a get request

paulinebanye: Thanks for the clarification, would adjust

backend/tests/endpoints/test_room.py
```diff
@@ -1,4 +1,132 @@
...
+        mock_get_user_room (AsyncMock): Asynchronous external api call
+        mock_dataStorage_read (AsyncMock): Asynchronous external api call
+    """
+    read_response = {"status_code": 424, "message": "Failed to retrieve room members"}
+
+    mock_get_user_room.return_value = fake_data_group_dm
+    mock_dataStorage_read.return_value = read_response
+
+    response = client.get(get_room_members_url)
+    assert response.status_code == 404
```
man-of-mind, line 131: the status_code ought to be 424 f

backend/tests/endpoints/test_room.py
```diff
@@ -1,4 +1,132 @@
...
+    assert response.json() == {"detail": "Room not found"}
```
man-of-mind, line 132: this should be Failure to retrieve room members

backend/tests/endpoints/test_room.py
```diff
@@ -1,4 +1,132 @@
...
+@pytest.mark.asyncio
+async def test_get_members_check_status_code(mock_get_user_room, mock_dataStorage_read):
+    """Get members unsuccessful when getting from zc core fails.
+    Args:
+        mock_get_user_room (AsyncMock): Asynchronous external api call
+        mock_dataStorage_read (AsyncMock): Asynchronous external api call
+    """
+    read_response = {"status_code": 424, "message": "Failed to retrieve room members"}
+
+    mock_get_user_room.return_value = fake_data_group_dm
```
man-of-mind, line 127: please pull and make the changes, this won't be needed again

backend/tests/endpoints/test_room.py
```diff
@@ -1,4 +1,132 @@
...
+        "detail": "Room not found",
+    }
+
+    response = client.get(get_room_members_url)
+    assert response.status_code == 404
+    assert response.json() == read_response
+
+
+@pytest.mark.asyncio
+async def test_get_members_reading_from_zc_core_returns_none(mock_get_user_room):
```
man-of-mind, line 106: this test case is the same as the one above

man-of-mind (commented):
@pauline-banye please go through the  comments

backend/tests/endpoints/test_room.py
paulinebanye: It failed with 424. The only thing that made it pass was 404

backend/conftest.py
```diff
@@ -49,3 +49,16 @@ def fixture_mock_centrifugo(mocker):
         side_effect=async_mock_centrifugo,
     )
     return async_mock_centrifugo
+
+
+@pytest.fixture(name="mock_dataStorage_read")
```
man-of-mind, line 54: This whole file have been moved inside the tests folder and you no longer need to write different fixtures for different endpoints.

man-of-mind (changes requested):
@pauline-banye please check the new comments.

backend/conftest.py
paulinebanye: Yeah I noticed that boss. Moved already

backend/tests/endpoints/test_room.py
paulinebanye: Done already. Just haven't pushed. Trying to fix and make them pass 🤦‍♀️

backend/tests/endpoints/test_room.py
```diff
@@ -1,4 +1,132 @@
...
 from main import app
 
 client = TestClient(app)
+
+
+get_room_members_url = (
+    "api/v1/org/6619ba4671a5f54782939d384//rooms/61dcf855eba8adb50ca13a24/members"
+)
+
+fake_data_group_dm = {
```
mikenrowland, line 12: @pauline-banye I think we should use a generic room data for tests concerning rooms instead of each person writing their own differently and for different room types, modifications can always be done to any field on the room object inside the test function when necessary to change the room data to suit that particular test case.

mikenrowland (commented):
@pauline-banye well done. I left a comment

PR #82, opened 2022-01-17
Get members test

paulinebanye:
Test for get room members successful.
Test for room not found.
Test for failure when retrieving room members


man-of-mind (approved):
Well done boss @pauline-banye

backend/tests/endpoints/test_room.py
```diff
@@ -1,4 +1,123 @@
...
 from fastapi.testclient import TestClient
 from main import app
+from utils.db import DataStorage
 
 client = TestClient(app)
+get_room_members_url = (
+    "api/v1/org/3467sd4671a5f5478df56u911/rooms/23dg67l0eba8adb50ca13a24/members"
+)
+
+fake_room_data = {
```
mikenrowland, line 13: well done @pauline-banye 
I think we need to agree on the fake room data structure, since we will be writing our tests in the same file, else our PRs will always be having conflicts.

mikenrowland (commented):
Hi @pauline-banye I left a comment

backend/tests/endpoints/test_room.py
paulinebanye: I agree @mykie88. Pls share ur thoughts re what should be adjusted.

What I actually used is the exact response returned when u create a room.  I only edited some values. 

I also took into account the tests for creating a room. One of the checks is a valid objid. So using any fake room without a valid objid throws an error.

mikenrowland: okay @pauline-banye , let's have a meet later and harmonize our thoughts. I will be free from 5pm

paulinebanye: A meet🤔. Aii. Buzz when you're free.

backend/tests/endpoints/test_room.py
```diff
@@ -1,4 +1,123 @@
...
+    assert response.status_code == 200
+    assert response.json() == read_response
+
+
+@pytest.mark.asyncio
+@mock.patch.object(DataStorage, "__init__", lambda x, y: None)
+async def test_get_members_room_not_found(mock_dataStorage_read):
+    """Tests when room is not found
+    Args:
+        mock_get_user_room (AsyncMock): Asynchronous external api call
```
zxenonx: It's rather `mock_dataStorage_read`

backend/tests/endpoints/test_room.py
```diff
@@ -1,4 +1,123 @@
...
+            "starred": False,
+        },
+        "619baa5c1a5f54782939d386": {
+            "closed": False,
+            "role": "member",
+            "starred": False,
+        },
+    }
+
+    read_response = {
```
zxenonx: What if we just named it `success_response`?

zxenonx, line 72: What if we just named it `success_response`?

zxenonx (commented):
Thanks @pauline-banye.

backend/tests/endpoints/test_room.py
paulinebanye: Oh, thanks for pointing that out. Wrote it before the change and I neglected to change it.

zxenonx (approved):
Thanks for the changes. @pauline-banye

PR #83, opened 2022-01-18
Sidebar fix

o4codes:
fixed display of name for personal DM


zxenonx (changes requested):
Thanks @o4codes. 

We also need to pop the member id when len(room_members) == 2 and room_type == DM.
This condition should take it into account as well.

o4codes:
When len(room_members) == 2 and room_type == DM, the member_id will be popped from the list of members.Do you get @zxenonx

backend/utils/sidebar.py
```diff
@@ -34,7 +34,9 @@ async def __get_room_members(
                             }
         """
         room_members = room.get("room_members")
-        room_members.pop(member_id, "not-found")  # remove self from room members
+
+        if len(list(room_members.keys())) > 1 and room.get("room_type") != RoomType.DM:
```
paulinebanye: @o4codes  You don't actually need .keys() again if you're using list. You can write it write is as list(room_members). Its the same thing.

This should also be 'and room.get("room_type") != RoomType.CHANNEL' so that it factors in GroupDMs and DMs
```suggestion
        if len(list(room_members)) > 1 and room.get("room_type") != RoomType.CHANNEL:
```

paulinebanye (changes requested):

backend/utils/sidebar.py
o4codes: Thanks
@pauline-banye  fixed

o4codes:
@zxenonx @pauline-banye  please go through the corrected reviews

mikenrowland (approved):
well done boss @o4codes

backend/utils/sidebar.py
```diff
@@ -34,7 +34,11 @@ async def __get_room_members(
                             }
         """
         room_members = room.get("room_members")
-        room_members.pop(member_id, "not-found")  # remove self from room members
+
+        if (
+            "".join(room_members.keys()) != member_id
+        ):  # checks if it's not a room with only the member_id
```
zxenonx: It works @o4codes.
However, i don't know if this part is easily understandable in terms of readability.
That's my opinion. Let's wait for other opinions.
@pauline-banye @mykie88

o4codes: I understand your view, 
I'll see if it can be made more readable.
Though I feel this is quite readable because it's just a check to see if a combination of all member_ids in a room is still equal to the member_id making the request, if it's true then the member_id is the only one in that room.

And there's a comment that describes it right there.
 @zxenonx

o4codes: @zxenonx checkout the commit

zxenonx: Thanks.
@o4codes

o4codes: Frankly I think the present condition we have covers fully for any room that is not a personal DM. 
Therefore, this is more readable than the former ones, and the condition is more extensive to cover for what is not a personal DM.

zxenonx (approved):
Thanks for the changes.

PR #84, opened 2022-01-20
Fix for room members endpoint and tests

paulinebanye:
- Fix for room members endpoint
- Fix get_room_members tests


zxenonx (approved):
Thanks. 
LGTM

PR #86, opened 2022-01-21
Fix error on update message endpoint

zxenonx:
Fixes https://linear.app/zurichat2/issue/ZUR-77/error-on-update-message-endpoint


backend/endpoints/messages.py
```diff
@@ -186,12 +186,13 @@ async def update_message(
...
 
-    message.update(payload)
-    message["edited"] = True
+    payload["edited"] = True
     edited_message = await DB.update(
-        MESSAGE_COLLECTION, document_id=message_id, data=message
+        MESSAGE_COLLECTION, document_id=message_id, data=payload
     )
 
+    message.update(payload)
```
paulinebanye, line 194: I think this should come before payload["edited"] = True

backend/endpoints/messages.py
```diff
@@ -186,12 +186,13 @@ async def update_message(
             detail="You are not authorized to edit this message",
         )
 
-    message.update(payload)
-    message["edited"] = True
+    payload["edited"] = True
```
paulinebanye, line 189:
```suggestion
    payload["edited"] = True
    message.update(payload)
```

backend/endpoints/messages.py
zxenonx: @pauline-banye, if we update the message object before setting the edited field to True in the payload, the edited field will still be False in the response.

paulinebanye: Ohhhh I forgot about the edited true flag 🤦‍♀️. 
Yeah it's needed beneath the edited true but why all the way underneath the await db.update?

zxenonx: Good point.
It's a common practice to declare/initialize variables as close to their usage as possible.
It's not a fixed rule though.
@pauline-banye

paulinebanye: Yeah for "readability" tho but it works and all the tests pass. Great job on the fix boss👍🏻

paulinebanye (approved):
Thanks for the fix @zxenonx

PR #87, opened 2022-01-22
Test for remove member endpoint

kinglighthill:
- Implemented tests for remove room member endpoint
- Modified rooms.py remove_member function. Added a None check to admin_id on line 130 because admin_id is optional
- Modified rooms_utils.py remove_room_member function. Changed the dict key from '_id' to 'id' on lines 150 and 160


backend/tests/endpoints/conftest.py
```diff
@@ -79,3 +79,47 @@ def fixture_mock_dataStorage_delete(mocker):
...
+
+
+@pytest.fixture(name="init_fake_room")
+def fixture_initialize_fake_room_data():
+    """A fixture that initialises a new fake room data for each test
+
+    Returns:
+        Dict: An room dictionary
+    """
+    room_data = {
```
paulinebanye: You can place a return here since you're simply returning the data for the room
```suggestion
    return {
```

backend/tests/endpoints/conftest.py
```diff
@@ -79,3 +79,47 @@ def fixture_mock_dataStorage_delete(mocker):
...
+        "created_at": "2022-01-11 03:18:02.364291",
+        "description": None,
+        "topic": "General Information",
+        "is_private": False,
+        "is_archived": False,
+        "id": "23dg67l0eba8adb50ca13a24",
+        "org_id": "3467sd4671a5f5478df56u911",
+        "created_by": "619ba4671a5f54782939d385",
+    }
+    return room_data
```
paulinebanye: Then return room data won't be needed
```suggestion
    
```

backend/.pre-commit-config.yaml
```diff
@@ -46,4 +46,6 @@ repos:
         - --disable=C0301
         - --disable=R0902
         - --disable=R0903
-        - --disable=C0103
\ No newline at end of file
+        - --disable=C0103
+        - --disable=R0913
+        - --disable=R0801
```
paulinebanye: What errors were u having here?

paulinebanye (changes requested):
This can be split in 2 PRs actually. The room utils and the room endpoint fixes can be in 1 PR. The conftest and test_rooms can be in another separate PR.

Well done on the tests @kinglighthill. It's no secret that I don't know how to write tests and I don't understand what parametrize does yet but all the tests passed so great job. 

Only thing I would say is except for the last two, the rest had the same name and I wasn't clear on what they were actually testing. If you could rename them to be more explicit, it would help.

backend/tests/endpoints/conftest.py
```diff
@@ -79,3 +79,47 @@ def fixture_mock_dataStorage_delete(mocker):
     zc_core_update_data = AsyncMock()
     mocker.patch("utils.db.DataStorage.delete", side_effect=zc_core_update_data)
     return zc_core_update_data
+
+
+@pytest.fixture(name="init_fake_room")
```
man-of-mind, line 94: I will suggest we create a separate file which will contains all the fake data, this will also reduce the number of arguments passed into the test functions.

backend/tests/endpoints/test_room.py
```diff
@@ -348,3 +373,152 @@ async def test_get_members_unsuccessful(mock_dataStorage_read):
...
+    response = client.patch(url=url)
+
+    assert response.status_code == status_code
+    assert response.json() == json_response
+
+
+@pytest.mark.asyncio
+@mock.patch.object(DataStorage, "__init__", lambda x, y: None)
+async def test_remove_room_member_wrong_room_id(
+    mock_dataStorage_read, mock_dataStorage_update
```
man-of-mind: The mock_update is not needed, the code won't reach the db.update when the room is empty.

backend/tests/endpoints/test_room.py
```diff
@@ -348,3 +373,152 @@ async def test_get_members_unsuccessful(mock_dataStorage_read):
...
+):
+    """
+    Leave room successfully
```
man-of-mind: Wrong description

backend/tests/endpoints/test_room.py
```diff
@@ -348,3 +373,152 @@ async def test_get_members_unsuccessful(mock_dataStorage_read):
...
+    response = client.patch(url=remove_room_member_url)
+
+    assert response.status_code == 404
+    assert response.json() == {"detail": "room does not exist"}
+
+
+@pytest.mark.asyncio
+@mock.patch.object(DataStorage, "__init__", lambda x, y: None)
+async def test_remove_room_member_from_dm(
+    init_fake_room, mock_dataStorage_read, mock_dataStorage_update
```
man-of-mind: mock_update not required as well

man-of-mind (changes requested):
Well done boss @kinglighthill 
I left some comments

backend/tests/endpoints/test_room.py
```diff
@@ -82,6 +82,31 @@
...
+
+fake_member_remove_room_member_url = (
+    "api/v1/org/3467sd4671a5f5478df56u911/rooms/23dg67l0eba8adb50ca13a24/"
+    + "members/1696f5ac4133ddaa309dcfe"
+)
+
+non_admin_remove_room_member_url = (
+    "api/v1/org/3467sd4671a5f5478df56u911/rooms/23dg67l0eba8adb50ca13a24/"
+    + "members/61696f5ac4133ddaa309dcfe?admin_id=619baa5c1a5f54782939d386"
+)
```
mikenrowland: too many urls for one endpoint @kinglighthill, if we go on creating a url for each test case we will end up having more urls than actual test codes.
Since the urls differ only slightly in characters, I suggest we use one url for each endpoint and use string methods to modify the characters within each test case when needed. This way we can reduce unnecessary repetitions. Thanks

backend/tests/endpoints/test_room.py
```diff
@@ -348,3 +373,152 @@ async def test_get_members_unsuccessful(mock_dataStorage_read):
...
+        (
+            f"{non_admin_remove_room_member_url}",
+            403,
+            {"detail": "must be an admin to remove member"},
+        ),
+    ],
+)
+@pytest.mark.asyncio
+@mock.patch.object(DataStorage, "__init__", lambda x, y: None)
+async def test_remove_room_member(
```
mikenrowland: omo! 6 parameters? do we really need to pass all these into the function?

backend/tests/endpoints/test_room.py
```diff
@@ -348,3 +373,152 @@ async def test_get_members_unsuccessful(mock_dataStorage_read):
     response = client.get(url=get_room_members_url)
     assert response.status_code == 424
     assert response.json() == {"detail": "Failure to retrieve room members"}
+
+
+@pytest.mark.parametrize(
```
mikenrowland: please can you elaborate on what this line is doing?

mikenrowland (commented):
well done @kinglighthill 
I left some comments.

backend/tests/endpoints/conftest.py
kinglighthill: Yes, please if you can it will be better. I created this fixture because I found out some tests modify the fake data declared at the beginning of the file

man-of-mind (commented):
Well done boss @kinglighthill you did a great job with the parameterize decorator, there is no test case when the endpoint will raise a connection error (you didn't check for instances when the update response will contain a different status code aside 200).

zxenonx:
I have resolved the conflicts on this PR.
So please @kinglighthill, check the **Allow edits by maintainers** option on this PR to let me push the changes.

kinglighthill:
> I have resolved the conflicts on this PR. So please @kinglighthill, check the **Allow edits by maintainers** option on this PR to let me push the changes.

It's checked already

zxenonx (approved):
LGTM.

PR #88, opened 2022-01-24
Reaction draft

paulinebanye:
Draft PR for adding and removing reactions.
it contains the merged add & remove endpoint and standalone add and remove reaction endpoints as well.
there's test utilizing only the add reaction endpoint.
@zxenonx @man-of-mind


backend/endpoints/messages.py
```diff
@@ -237,3 +238,508 @@ async def get_messages(org_id, room_id):
         content=ResponseModel.success(data=response, message="Messages retrieved"),
         status_code=status.HTTP_200_OK,
     )
+
+
+@router.put(  # fully functional (merged) add and remove reactions
```
zxenonx: @pauline-banye, 😅  this fully merged add & remove endpoints though...
Before we move on to the tests you wrote, we first need to split these endpoints into smaller functions.
As you can see both add and remove endpoints are really big. As a result, they won't be easily readable and more importantly, it will be difficult to write tests for them.
As Emoji is a resource by itself also, I suggest creating an emoji_utils file for helper functions related to reactions, so that the endpoints functions are more concise and smaller...
If we do it this way, the tests will be easier to write.

zxenonx (commented):
@pauline-banye, I left a comment. Thanks.

backend/endpoints/messages.py
paulinebanye: @zxenonx yeah that merged endpoint is massive. That's actually y I added a separate add and remove endpoint as well. You didn't look @ it? 🤔. 

It's even outlined it in the PR description

paulinebanye: Re adding an emoji_utils file, is it actually relevant? 

The only use case I can think of for it is probably to retrieve all the emojis for a certain message. If we need to, we can put that in the message_utils.

paulinebanye: The tests I've written so far are just for the add_reaction endpoint. 

I've actually made some changes to the endpoint, let me push so u can see the updates.

zxenonx: > @zxenonx yeah that merged endpoint is massive. That's actually y I added a separate add and remove endpoint as well. You didn't look @ it? thinking.
> 
> It's even outlined it in the PR description

I did look at it. But they are still very massive.

zxenonx: > 
> 
> Re adding an emoji_utils file, is it actually relevant?
> 
> The only use case I can think of for it is probably to retrieve all the emojis for a certain message. If we need to, we can put that in the message_utils.
> 

In the endpoints, we will only have to call the helper functions we will have created in the utils file. This will reduce the size of the endpoint functions considerably and make unit testing easier.

zxenonx: > The tests I've written so far are just for the add_reaction endpoint.
> 
> I've actually made some changes to the endpoint, let me push so u can see the updates.

But we have to keep in mind that we will only have to expose one endpoint to handle adding and removing an emoji, right?

paulinebanye: That's y i did a merged endpoint initially. But if we don't use the merged, it would be 2 endpoints

paulinebanye: > 

Which helper function do we need for the emoji?

zxenonx: > That's y i did a merged endpoint initially. But if we don't use the merged, it would be 2 endpoints

Yeah but will be easier with something like this :
```python
async def reactions () :
  ....
  if blabla:
     add_reaction()
  else:
      remove_reaction() 
```

zxenonx: > > 
> 
> Which helper function do we need for the emoji?

The helper functions will actually do the manipulations to add/remove a reaction. 
Then they will be called in the endpoint functions...

paulinebanye: I think I need to go over the endpoint with u bcos the only thing that adds and removes is a line saying either append or remove.

And then after that is the update database and publish. That's actually what's taking up the bulk of the endpoint.

So writing a helper function to address that one line doesn't seem relevant to me.

PR #93, opened 2022-02-09
Refactor message_utils.py

zxenonx:
Fixes https://linear.app/zurichat2/issue/ZUR-79/refactor-message-utilspy


backend/utils/message_utils.py
```diff
@@ -1,52 +1,118 @@
...
+
     DB = DataStorage(org_id)
     query = {"room_id": room_id, "_id": message_id}
-    response = await DB.read(MESSAGE_COLLECTION, query=query)
-    if response and "status_code" not in response:
-        return response
-    return {}
+    response = await DB.read(settings.MESSAGE_COLLECTION, query=query)
+
+    if not response or "status_code" in response:
```
o4codes, line 115: This looks great, But without the knowledge of operator precedence, can be quite confusing to the reader on first glance.
If it can be rewritten as 
 
`if (not response) or ("status_code" in response)`

I feel this might make easy to grasp on first glance.

o4codes (commented):
Good work @zxenonx 
The docs are on fireeeeee.
I left some remarks

backend/utils/message_utils.py
```diff
@@ -1,52 +1,118 @@
...
+async def get_message(
+    org_id: str, room_id: str, message_id: str
+) -> Optional[dict[str, Any]]:
+    """Get a specific message in a room.
+
     Args:
         org_id (str): The organization id
-        room_id (str): The room id
-        message_id (str): The message id
+        room_id (str): The id of the room where to the message is to be retrieved.
```
paulinebanye:
```suggestion
        room_id (str): The id of the room where the message is to be retrieved.
```

backend/utils/message_utils.py
zxenonx: Thanks, @o4codes, It's a good suggestion but it keeps me thinking.

As you have pointed out, the real issue here is the reader not having much knowledge of operator precedence.
Because there's no change in the op precedence in the expr,  `if not response or "status_code" in response` and `if (not response) or ("status_code" in response)` will be evaluated the same.

I would definitely have added some parentheses if the expr was a bit more tricky or longer, or if it contained an additional operator for example.

As we do not want the expr to be evaluated as ` if (not response or "status_code") in response`, I think it's ok.
If we add parentheses, are we really helping the reader? Probably.
But wouldn't it be better to refer them to an article about how operator precedence works instead?

Maybe I'm biased... Let me know what you think about this.

o4codes: I don't think you are biased. And yea `if (not response) or ("status_code" in response)` and `if not response or status_code in response` would still produce the same output due to the operator precedence. 

Just anyone without the knowledge of operator precedence could easily read this as `if not(response or status_code in response)`, and therefore attempt to fix what is not broken. I only felt we could make this easier to for the reader to understand at first glance. 

And yea, we can leave it as it is, and probably have an article for anyone who doesn't understand it. But then we might not be always here in this project though to resolve issues like this.

@zxenonx

paulinebanye (approved):
@zxenonx 👍🏻

backend/utils/message_utils.py
zxenonx: Anyone who wants to fix something in the code, must first make sure the code is not doing what it is supposed to do, right?

Please, let's leave it as it is.
@o4codes

PR #96, opened 2022-02-10
Updated remove member

kinglighthill:
Fixed a crash in the remove_member endpoint
Parsed ValueError and ConnectionError successfully to HTTPException


backend/endpoints/rooms.py
```diff
@@ -162,12 +161,12 @@ async def remove_member(
...
     except ValueError as value_error:
         raise HTTPException(
-            detail=value_error, status_code=status.HTTP_404_NOT_FOUND
+            detail=str(value_error), status_code=status.HTTP_404_NOT_FOUND
         ) from value_error
 
     except ConnectionError as connect_error:
         raise HTTPException(
-            detail=json.dumps(str(connect_error)),
+            detail=str(connect_error),
```
o4codes, line 169: Ohhh 
The exception objects are not JSON encodable
Nice work

o4codes (approved):
Good work

mikenrowland (commented):
well done @kinglighthill good job.

I don't know why but for some reason this endpoint has turned out to be unnecessarily complicated. We might have to sit down in the near future and look at how we can simplify the design.

backend/endpoints/rooms.py
```diff
@@ -132,11 +131,10 @@ async def remove_member(
             status_code=status.HTTP_404_NOT_FOUND,
             detail="admin id specified not a member of the room",
         )
-    
+
     if admin_id is not None and member_id == admin_id:
         raise HTTPException(
-            status_code=status.HTTP_403_FORBIDDEN,
-            detail="cannot remove yourself"
+            status_code=status.HTTP_403_FORBIDDEN, detail="cannot remove yourself"
```
paulinebanye, line 137: Based on slack behavior and the compiled zurichat document, there are no restrictions for members leaving a channel. A public channel can be completely empty.

The only restriction for leaving channels is for an empty private channel. And in that case the last member is prompted to archive the channel. The last member doesn't have to be the creator or admin.

I think this endpoint needs to be reworked to reflect that behavior.

paulinebanye (changes requested):

backend/endpoints/rooms.py
kinglighthill: I'm not restricting anyone from leaving the channel, members can still leave the channel. Due to the nature of the endpoitn with is solving two problems leaving a channel now has 2 ways. I'm just making the consumer of the endpoint to just one way of implementing leaving a channel.

zxenonx (approved):

PR #137, opened 2022-11-24
feat(BE-50): implementing pagination before filtering

bolexs:
**FEAT(BE-50)** - [linear issue link](https://linear.app/team-gear/issue/BAC-50/zuri) 
Pagination - implemented zuri chat pagination for dm and channel with @Shopiley
@zxenonx @mykie88

### DIFF
--- a/backend/endpoints/messages.py
+++ b/backend/endpoints/messages.py
@@ -1,4 +1,5 @@
 from fastapi import APIRouter, BackgroundTasks, HTTPException, status
+from fastapi_pagination import Page, add_pagination, paginate
 from schema.message import Message, MessageRequest
 from schema.response import ResponseModel
 from starlette.responses import JSONResponse
@@ -215,7 +216,7 @@ async def update_message(
 
 @router.get(
     "/org/{org_id}/rooms/{room_id}/messages",
-    response_model=ResponseModel,
+    response_model=Page[Message],
     status_code=status.HTTP_200_OK,
     responses={424: {"detail": "ZC Core failed"}},
 )
@@ -270,7 +271,6 @@ async def get_messages(org_id: str, room_id: str):
             detail="Zc Core failed",
         )
 
-    return JSONResponse(
-        content=ResponseModel.success(data=response, message="Messages retrieved"),
-        status_code=status.HTTP_200_OK,
-    )
+    return paginate(response)
+
+add_pagination(router)
--- a/backend/requirements.txt
+++ b/backend/requirements.txt
@@ -17,7 +17,7 @@ fastapi==0.70.0
 fastapi-pagination==0.9.1
 filelock==3.3.1
 h11==0.12.0
-httpcore==0.13.7
+httpcore==0.15.0
 httptools==0.2.0
 httpx==0.23.0
 identify==2.3.1


backend/endpoints/messages.py
```diff
@@ -270,7 +271,6 @@ async def get_messages(org_id: str, room_id: str):
             detail="Zc Core failed",
         )
 
-    return JSONResponse(
-        content=ResponseModel.success(data=response, message="Messages retrieved"),
-        status_code=status.HTTP_200_OK,
-    )
+    return paginate(response)
+
```
AI-fae, line 275: you need to return a status code and message along with this data.

AI-fae (changes requested):

FlorenceEgwu:
Nice work.
But this solution is not scalable because the pagination is done after querying the database. 
This becomes a problem as the data grows bigger say 10k+ (which we are likely to get in the shortest time). By this time, we would be fetching all 10k data into memory for each call before paginating. 

Note: 
this is no fault of yours. If anything is to be faulted, it is the pagination package itself.

solution:
Write a custom pagination query. sounds daunting at first but is pretty straightforward.

If you are interested in working on a more scalable solution, you can reach out to me let's discuss.

FlorenceEgwu:
you can look at the [bigfastapi repository ](https://github.com/bigfastcode/bigfastapi/blob/staging/bigfastapi/utils/paginator.py) for inspiration on how to implement this.

backend/endpoints/messages.py
```diff
@@ -1,4 +1,5 @@
 from fastapi import APIRouter, BackgroundTasks, HTTPException, status
+from fastapi_pagination import Page, add_pagination, paginate
```
mikenrowland, line 2: @bolexs how are you using paginate? What's the behaviour and did you test this implementation with data from the DB? I am asking because I can see a caveat in the [documentation](https://pypi.org/project/fastapi-pagination/), that warns against using the default paginate method when working with databases.

mikenrowland (commented):
@bolexs I left a comment

bolexs:
> Nice work. But this solution is not scalable because the pagination is done after querying the database. This becomes a problem as the data grows bigger say 10k+ (which we are likely to get in the shortest time). By this time, we would be fetching all 10k data into memory for each call before paginating.
> 
> Note: this is no fault of yours. If anything is to be faulted, it is the pagination package itself.
> 
> solution: Write a custom pagination query. sounds daunting at first but is pretty straightforward.
> 
> If you are interested in working on a more scalable solution, you can reach out to me let's discuss.

Thank you, i'm working on this and will get this done

bolexs:
> you can look at the [bigfastapi repository ](https://github.com/bigfastcode/bigfastapi/blob/staging/bigfastapi/utils/paginator.py) for inspiration on how to implement this.

thank you, i will look into it now

backend/endpoints/messages.py
AI-fae: okay, I see why you couldn't do that

FlorenceEgwu (approved):

FlorenceEgwu (approved):

mikenrowland (approved):
We can merge and keep working on making this better

zxenonx (approved):
Good work @bolexs.

PR #139, opened 2022-11-25
BAC-16 Create Endpoint for creating and updating Thread Messages

rector619:
https://linear.app/team-powerdrill/issue/BAC-16/create-end-point-for-creating-and-updating-thread-messages

### DIFF
--- a/backend/endpoints/threads.py
+++ b/backend/endpoints/threads.py
@@ -11,7 +11,7 @@ import json
 router = APIRouter()
 
 @router.put(
-		"/org/{org_id}/rooms/{room_id}/messages/{message_id}/edithreads",
+		"/org/{org_id}/rooms/{room_id}/messages/{message_id}/threads",
 		response_model=ResponseModel,
 		status_code=status.HTTP_200_OK,
 		responses={
@@ -86,7 +86,6 @@ async def update_threads_message(
 				HTTPException [424]: Message not edited.
 		"""
 
-		# payload = Message(**request.dict(), org_id=org_id, room_id=room_id)
 		payload = request.dict()
 
 		loadedMessage = await get_message(org_id, room_id, message_id)
@@ -100,11 +99,6 @@ async def update_threads_message(
 						status_code=status.HTTP_404_NOT_FOUND, detail="Message not found"
 				)
 		
-		# if message["sender_id"] != payload["sender_id"]:
-		#     raise HTTPException(
-		#         status_code=status.HTTP_401_UNAUTHORIZED,
-		#         detail="You are not authorized to edit this message",
-		#     )
 
 		added_thread_message = await edit_message_threads(org_id, message_id, thread_to_add)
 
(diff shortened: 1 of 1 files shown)


backend/endpoints/threads.py
```diff
@@ -1,3 +1,131 @@
...
+from utils.centrifugo import Events, centrifugo_client
+from utils.message_utils import get_message, get_room_messages
+from utils.message_utils import update_message_threads as edit_message_threads
+from utils.room_utils import get_org_rooms
+import json
 
 router = APIRouter()
+
+@router.put(
+		"/org/{org_id}/rooms/{room_id}/messages/{message_id}/edithreads",
```
mikengr: rename `edithreads` to just `thread`

backend/endpoints/threads.py
```diff
@@ -1,3 +1,131 @@
...
+		if not loadedMessage:
+				raise HTTPException(
+						status_code=status.HTTP_404_NOT_FOUND, detail="Message not found"
+				)
+		
+		# if message["sender_id"] != payload["sender_id"]:
+		#     raise HTTPException(
+		#         status_code=status.HTTP_401_UNAUTHORIZED,
+		#         detail="You are not authorized to edit this message",
+		#     )
```
mikengr: why is your PR filled with commented code?

backend/endpoints/threads.py
```diff
@@ -1,3 +1,131 @@
...
+
+		added_thread_message = await edit_message_threads(org_id, message_id, thread_to_add)
```
mikengr, line 130: The endpoint is `update_threads_message` what is this line doing?

backend/endpoints/threads.py
```diff
@@ -1,3 +1,131 @@
...
+		# Publish to centrifugo in the background.
+		background_tasks.add_task(
+				centrifugo_client.publish,
+				room_id, 
+				Events.MESSAGE_UPDATE, 
+				loadedMessage
+		)
+
+		return JSONResponse(
+				content=ResponseModel.success(data=loadedMessage, message="Thread Added"),
```
mikengr, line 150: I am not sure this endpoint is doing what you say it is supposed to do.

mikengr (changes requested):
@RECTOR619 I left some comments

backend/endpoints/threads.py
rector619: i  will change it now sir

backend/endpoints/threads.py
rector619: it was a mistake i forgot to clear it before pushing.........corrected sir

backend/endpoints/threads.py
rector619: my endpoint is creating a thread under a specified message like the way slack does, you can decided to reply someone message using thread so that is why am using put endpoint because am trying to embed a new message inside an old message which will form a thread .........then i now create a function that you can use to update it

backend/endpoints/threads.py
rector619: the endpoint is to create a thread and update it

PR #141, opened 2022-11-25
created a script that triggers message notification for users in a channel and for DM

matthew-akinola:
**BAC-26** (https://linear.app/team-axe/issue/BAC-26/trigger-messaging-events-step-3)
The chatNotificationTrigger script:

- create a Novu subscription for users
- create a notification trigger for users in DM, Channels or Group DM
- create a notification trigger for tagged users 
Task Info:

- to create Novu subscription for DM members, import the **"Notification"** class into the **"utils/room_utils"** file, in the "create_room" function, instantiate the **"Notification.dm_subscriber_create"** and pass the **"room_obj"** as an argument.
- to create Novu subscription for channel members, import the **"Notification"** class into the **"utils/room_utils"**, in the "join_room" function, instantiate the **"Notification.channel_subscriber_create"** and pass the "member" variable as an argument.
- - to instantiate  Novu  message notification function,  import the "Notification" class into the **"endpoint/messages"**,  instantiate the **"Notification.channel_messages_trigger"** and pass the "message" variable as an argument.

### DIFF
--- a/backend/utils/chat_notification.py
+++ b/backend/utils/chat_notification.py
@@ -223,7 +223,7 @@ class Notification:
             room_member_list.append(member_id)
          # send a message notification to every user in either in the channel
          #  or Group DM by calling the Novu trigger method
-        channel_or_group_msg_trigger = await event.trigger(
+        channel_or_group_msg_notification = await event.trigger(
             '<REPLACE_WITH_EVENT_NAME_FROM_ADMIN_PANEL>',
             {
                 "to": room_member_list,
@@ -231,7 +231,7 @@ class Notification:
             })
         # raise an http exception if novu fails to send message notification
         # to group DM or channel room members
-        if channel_or_group_msg_trigger["acknowledged"] != "true":
+        if channel_or_group_msg_notification["acknowledged"] != "true":
             raise HTTPException(
                 status_code=422, 
                 detail="Message notification failed"
@@ -251,7 +251,7 @@ class Notification:
                     status_code=422, 
                     detail="failed to create a DM notification"
                     )
-        return channel_or_group_msg_trigger
+        return channel_or_group_msg_notification
         
 
 
--- a/backend/utils/chat_notification.py
+++ b/backend/utils/chat_notification.py
@@ -27,23 +27,25 @@ class Notification:
             HTTP_422- when Novu couldn't send notification to tagged users
         """
         tagged_users_list = []
-        user_msg_tag = message_obj.get("entityMap", " ")
+        get_tagged_users = message_obj.get("richUiData", " ")
         if user_msg_tag:
-            for message in range(len(user_msg_tag)):
-                member_id = user_msg_tag[str(message)]['data']['mention']['member_id']
-                tagged_users_list.append(member_id)
-            tagged_users = await event.trigger(
-                'you-have-been-mentioned',
-                {
-                    "to": tagged_users_list,
-                    "payload": payload
-                })
-            if tagged_users['statusCode'] !=201:
-                raise HTTPException(
-                    status_code=422, 
-                    detail="Novu couldn't send notifications to tagged users"
-                    )
-            return tagged_users
+            user_msg_tag = get_tagged_users['entityMap']
+            if user_msg_tag !=[]:
+                for message in range(len(user_msg_tag)):
+                    member_id = user_msg_tag[str(message)]['data']['mention']['name']
+                    tagged_users_list.append(member_id)
+                tagged_users = await event.trigger(
+                    'you-have-been-mentioned',
+                    {
+                        "to": tagged_users_list,
+                        "payload": payload
+                    })
+                if tagged_users['statusCode'] !=201:
+                    raise HTTPException(
+                        status_code=422, 
+                        detail="Novu couldn't send notifications to tagged users"
+                        )
+                return tagged_users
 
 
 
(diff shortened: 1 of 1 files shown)


backend/utils/chat_notification.py
```diff
@@ -0,0 +1,218 @@
...
+            payload-> Dict: message data that should be passed to novu
+            sender_id -> str
+        Raise:
+            HTTP_422- when Novu couldn't send notification to tagged users
+        """
+        tagged_users = []
+        # user_msg_tag = message_obj["richUiData"]['entityMap']
+        user_msg_tag = message_obj.get("entityMap", " ")
+        if user_msg_tag:
+            for i in range(len(user_msg_tag)):
```
mikengr: you can't use `'i'` to name a variable in a codebase like this, it is confusing and can be misleading. I suggest you use a proper variable name that clearly represents the data held by this variable.

backend/utils/chat_notification.py
```diff
@@ -0,0 +1,218 @@
...
+                {
+                    "to": tagged_users,
+                    "payload": payload
+                })
+            if dm_trigger_create['statusCode'] !=201:
+                raise HTTPException(
+                    status_code=422, 
+                    detail="Novu couldn't send notifications to tagged users"
+                    )
+            return dm_trigger_create
```
mikengr: `dm_trigger_create`??? Can you explain? the function says `check_message_tag` . What's the return type of this function?

backend/utils/chat_notification.py
```diff
@@ -0,0 +1,218 @@
...
+            if create_subcriber['statusCode'] !=201:
+                raise HTTPException(
+                status_code=422, 
+                detail="Novu couldn't create a Novu subscription for Novu"
+                )
+        return create_subcriber
+        
+        
+
+    async def channel_subcriber_create(self, member_id):
```
mikengr, line 100: how do you know what channel or room to subscribe the user/member to?

backend/utils/chat_notification.py
```diff
@@ -0,0 +1,218 @@
...
+        try:
+            message = message_obj['data']['blocks'][0]['text']
+            sender_id = message_obj.get("sender_id")
+        except:
+            pass
+        room = get_room(org_id, room_id)
+        if not room:
+            return ValueError("room with ID not found")
+        # create a notfication for the DM user 
+        if room['roomt_type'] == 'DM':
```
mikengr: This function will break here '`roomt_type`'

backend/utils/chat_notification.py
```diff
@@ -0,0 +1,218 @@
...
+                raise HTTPException(
+                    status_code=422, 
+                    detail="Failed to process Novu instance"
+                    )
+            if dm_create["acknowledged"] != "true":
+                raise HTTPException(
+                status_code=422, 
+                detail="failed to create a DM Novu instance"
+                )
+            return "dm notification trigger successful"
```
mikengr, line 215: the function says `channel_messages_trigger`, why are we handling dm notifications here?

mikengr (changes requested):
Good work @matthew-akinola . I left some comments.

backend/utils/chat_notification.py
matthew-akinola: everything fixed now boss @mikengr

backend/utils/chat_notification.py
matthew-akinola: when a message object is created, the function differentiates the messages and then call either the method responsible for DM notification(trigger) or channel. I've changed the name @mikengr

backend/utils/chat_notification.py
matthew-akinola: The purpose of this task is to subscribe users on the Novu DB. since every room has members, I get the room ID from the message object, gets the members which definitely would have become a Novu subscriber either via DM or joining a channel and then send notification to this users. @mikengr

backend/utils/chat_notification.py
```diff
@@ -216,13 +216,13 @@ async def messages_trigger(self, message_obj):
...
-        # send a message notification to every user in a room by 
-        # calling the Novu trigger method
         get_members = await get_room_members(org_id, room_id)
         for member_id, value in get_members.items():
             if member_id == sender_id:
                 del get_members[member_id]
             room_member_list.append(member_id)
+         # send a message notification to every user in either in the channel
+         #  or Group DM by calling the Novu trigger method
         channel_or_group_msg_trigger_crate = await event.trigger(
```
mikengr: channel_or_group_msg_trigger_`crate`?

backend/utils/chat_notification.py
mikengr: This sounds like an assumption... `which definitely would have`??

matthew-akinola: Not an assumption, maybe I picked the wrong choice of word but what I'm insinuating is, you'll become a Novu subscriber while trying to join a room. All that's needed is to send message to members in a room since each room as an ID and respective members @mikengr

backend/utils/chat_notification.py
matthew-akinola: fixed @mikengr

backend/utils/chat_notification.py
matthew-akinola: fixed @mikengr

backend/utils/chat_notification.py
matthew-akinola: fixed @mikengr

mikengr (approved):

PR #142, opened 2022-11-25
Feat/bac 70 build update channel endpoint

nodebe:
- Gets the required data from the client-side(room_id, org_id, member_id, data to update)
- Checks if member_id is a member of room
- Checks if member_id is an admin to update room
- Updates room and returns updated data

[Link to Linear Issue](https://linear.app/team-bevel/issue/BAC-70/build-an-endpoint-to-update-channel-zuri-chat-task)

### DIFF
--- a/backend/endpoints/rooms.py
+++ b/backend/endpoints/rooms.py
@@ -3,7 +3,7 @@ from typing import Dict, Optional
 from fastapi import APIRouter, BackgroundTasks, Body, HTTPException, status
 from fastapi.responses import JSONResponse
 from schema.response import ResponseModel
-from schema.room import Role, Room, RoomMember, RoomRequest, RoomType
+from schema.room import Role, Room, RoomMember, RoomRequest, RoomType, UpdateRoomRequest
 from utils.centrifugo import Events, centrifugo_client
 from utils.db import DataStorage
 from utils.room_utils import get_room, remove_room_member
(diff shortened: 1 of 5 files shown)


backend/endpoints/rooms.py
```diff
@@ -435,3 +435,76 @@ async def get_members(org_id: str, room_id: str):
...
+    )
+
+    if updater.get("role") != Role.ADMIN:
+        raise HTTPException(
+        status_code=status.HTTP_401_UNAUTHORIZED,
+        detail="must be an admin to update room.",
+    )
+
+    # Insert members into data request to avoid deletion of all members
+    data['members'] = members
```
mikengr, line 491: Why do you need this? A patch update only modifies specified fields. How do you mean to avoid deletion of all members?

mikengr (commented):
@nodebe good work. I left a comment

backend/endpoints/rooms.py
nodebe: For the MongoDB collection, passing in a Json makes it easier to update. If the line of code is not there, the Json will be used to replace the current on it. 

It's easier to request all and update straight up.

mikengr: I still don't understand your explanation. @nodebe

nodebe: Okay. 
The document is in Json format. It is easier to pass it generally than individually. 

During the edit, members will not be edited so it's not necessary to pass the data along. 

That's why I fetched the members and reattached to the data after it was passed.

backend/endpoints/rooms.py
```diff
@@ -435,3 +435,76 @@ async def get_members(org_id: str, room_id: str):
...
+
+    # updates the room data in the db collection
+    update_response = await DB.update(
+        settings.ROOM_COLLECTION, document_id=room_id, data=data
+    )  
+
+    if update_response and update_response.get("status_code", None) is None:
+        room_id = {"room_id": update_response.get("data").get("object_id")}
+
```
FlorenceEgwu: why do you need to change the room id. shouldn't it be the same as the one originally fetched?

FlorenceEgwu (approved):

backend/endpoints/rooms.py
nodebe: That's a honest mistake. I'm suppose to remove that line. Thank you for pointing it out

PR #152, opened 2022-11-25
fix(BAC-49):implemented custom pagination

Shopiley:
Task: 
Implement progressive requesting of data and old chats so it does not load all of history. This goes for DMs and Channel Chats. (Zuri Project 2)

[Linear link](https://linear.app/team-gear/issue/BAC-49/zuri-chat-task)

Solution:
Implemented a custom pagination query to get data in batches (of 15) from the database with @bolexs.

Result:
image
(note|: limit and page are optional query parameters; when not specified page=1 and limit=15)

**New Edit:
The parameters are now page & size not page & limit**

image
image

Please review:
@zxenonx @AI-fae @mykie88

### DIFF
--- a/backend/endpoints/messages.py
+++ b/backend/endpoints/messages.py
@@ -272,11 +272,12 @@ async def get_messages(org_id: str, room_id: str, page: int = 1, limit: int = 15
         )
 
     result = {
-            "status": "success",
-            "message": "Messages retrieved",
             "data": response,
             "page": page,
             "size": limit
     }
-    return result
 
+    return JSONResponse(
+        content=ResponseModel.success(data=result, message="Messages retrieved"),
+        status_code=status.HTTP_200_OK,
+    )
--- a/backend/endpoints/messages.py
+++ b/backend/endpoints/messages.py
@@ -216,11 +216,11 @@ async def update_message(
 
 @router.get(
     "/org/{org_id}/rooms/{room_id}/messages",
-    response_model=Page[Message],
+    response_model=[].append(Message),
     status_code=status.HTTP_200_OK,
     responses={424: {"detail": "ZC Core failed"}},
 )
-async def get_messages(org_id: str, room_id: str):
+async def get_messages(org_id: str, room_id: str, page: int = 1, limit: int = 15):
     """Fetches all messages sent in a particular room.
 
     Args:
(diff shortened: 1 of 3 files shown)


backend/endpoints/messages.py
```diff
@@ -216,11 +216,11 @@ async def update_message(
 
 @router.get(
     "/org/{org_id}/rooms/{room_id}/messages",
-    response_model=Page[Message],
+    response_model=[].append(Message),
```
FlorenceEgwu, line 219: what's going on here? 😂
Please use the custom ResponseModel (the one you removed in your solution).
use do `List[Messages]`.

I am not comfortable with how this looks

backend/utils/message_utils.py
```diff
@@ -81,7 +86,7 @@ async def get_room_messages(
 
 
 async def get_message(
-    org_id: str, room_id: str, message_id: str
+    org_id: str, room_id: str, message_id: str, page: int, limit: int
 ) -> Optional[dict[str, Any]]:
     """Get a specific message in a room.
```
FlorenceEgwu: pagination is not needed in this function. since it gets a single message.

backend/utils/paginator.py
```diff
@@ -0,0 +1,2 @@
+async def off_set(page: int, limit: int):
+    return (page-1)*limit
```
FlorenceEgwu, line 2: nice work

backend/endpoints/messages.py
```diff
@@ -271,6 +271,4 @@ async def get_messages(org_id: str, room_id: str):
             detail="Zc Core failed",
         )
 
-    return paginate(response)
-
-add_pagination(router)
+    return response
```
FlorenceEgwu: you need to return some other information with this data to enable frontend implementation.
e.g
1. current page (also known as page)
2. size of returned data
3. total number of items in db
4. pointer to the next page
5. pointer to the previous page

FlorenceEgwu (dismissed):

FlorenceEgwu:
Well done guys.
Happy to see that you guys went through with in this.

I have added some review comments

backend/endpoints/messages.py
bolexs: > Well done guys. Happy to see that you guys went through with in this.
> 
> I have added some review comments

thank you, we would work on your comments

backend/endpoints/messages.py
mikengr: @Shopiley don't forget this

mikengr (approved):
LGTM

backend/utils/message_utils.py
Shopiley: Oh right, I'll remove it.

backend/endpoints/messages.py
Shopiley: Yeah, I have made another [pull request](https://github.com/zurichat/zc_messaging/pull/165#issue-1467048882) to effect these changes.

PR #157, opened 2022-11-26
Feat/bac 55 adding upload function data storage

devvspaces:
- Created file storage function to upload files to zc_core
- Handled all error edge cases
- Wrote tests for all cases

Linear: https://linear.app/team-socket/issue/BAC-55/implementing-the-data-storage-upload-function-to-upload-one-or-more

pr


backend/requirements.txt
```diff
@@ -18,7 +18,7 @@ fastapi-pagination==0.9.1
 filelock==3.3.1
 h11==0.12.0
 httpcore==0.15.0
-httptools==0.2.0
+httptools==0.5.0
```
FlorenceEgwu: take out the versioning on this  completely. I think this still gives issue for some people.
i.e we should have
```suggestion
httptools
```

backend/tests/utils/test_db.py
```diff
@@ -14,3 +16,89 @@ async def test_connection_timeout():
...
+                "data": "test"
+            }
+            response = await TestDB.upload(files, token)
+            assert response == {
+                "status_code": 400,
+                "message": {
+                    "status": 400,
+                    "data": "test"
+                }
+            }
```
FlorenceEgwu: great job on adding these test cases. Really happy with it.

backend/utils/db.py
```diff
@@ -373,3 +377,59 @@ async def get_member(
...
+                }
+        """
+
+        files = [("files", file) for file in files]
+        url = self.upload_api + f"/{self.plugin_id}"
+
+        headers = {
+            "Authorization": token,
+        }
+
```
FlorenceEgwu: I thought headers like this were handled somewhere central in this class.
@zxenonx , can you remember?

backend/utils/db.py
```diff
@@ -373,3 +377,59 @@ async def get_member(
         if members:
             for member in members:
                 return member if member["_id"] == member_id else {}
+
+    async def upload(
+        self, files: List[Any], token: str
```
FlorenceEgwu: `files_upload` would be a more descriptive name for this method

FlorenceEgwu (dismissed):

FlorenceEgwu:
Great job on this.
I have added a few comments

backend/utils/db.py
devvspaces: Okay thanks, I will make the required changes

backend/utils/db.py
zxenonx: @AI-fae, I remember we did this in zc_plugin_dm but not here.
Besides, we need to protect these endpoints now.

backend/utils/message_utils.py
```diff
@@ -1,4 +1,4 @@
-from typing import Any, Optional, List, Dict
+from typing import Any, Dict, List, Optional
```
zxenonx: We no longer need to import List and Dict from Python 3.9.
https://peps.python.org/pep-0585/
list and dict can be used. They are built in

zxenonx (approved):
Good job

backend/utils/message_utils.py
zxenonx: @devvspaces

mikengr (approved):
LGTM

PR #140, opened 2022-11-25
BAC-25(https://linear.app/team-axe/issue/BAC-25/python-sdk-step-2)

Huzzy619:
NovuPy (_Python SDK for Novu_) covers major endpoints and functionalities of the Novu Notification system.
[Link to Linear](https://linear.app/team-axe/issue/BAC-25/python-sdk-step-2)
- Events
- Subscribers
- Activity
- Integrations
- Feeds  

All others can be configured from the admin page of the Novu Instance

The url of the hosted Novu Instance can be configured in the settings module.
The `NOVU_API_KEY` should be inside a `.env` file

As of this moment, no live link has been provided to me. 
But the SDK can still be tested by calling the function and passing required data.

For additional information on how to use the NovuPy  [click here](https://docs.novu.co/api/trigger-event/)

PR #159, opened 2022-11-27
Fix/bac 25 setup python novu sdk

Huzzy619:
Added the live Novu instance url and credentials

### DIFF
--- a/backend/NovuPy/activity.py
+++ b/backend/NovuPy/activity.py
@@ -1,0 +1,45 @@
+import httpx
+
+from .settings import Core
+
+
+class Activity(Core):
+
+    async def get_activity(self, **kwargs):
+        """
+        Get activity feed
+
+        A page query parameter can be passed into the function to paginate response e.g page = 2  
+        Example response:
+
+        {
+            totalCount: 0,
+            data: ["data"],
+            pageSize: 0,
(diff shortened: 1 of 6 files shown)


backend/NovuPy/activity.py
```diff
@@ -0,0 +1,45 @@
...
+            weeklySent: 0,
+            monthlySent: 0,
+            yearlySent: 0
+        }
+        """
+        url = self.base_url+'v1/activity/stats'
+        async with httpx.AsyncClient() as requests:
+            response = await requests.get(url=url, headers=self.s_header)
+
+        return response.json()
```
mikengr, line 45: why is this code repeated? same here
#140

mikengr (commented):
Everything in this PR is the same as #140 , why is that?

backend/NovuPy/activity.py
Huzzy619: Actually what happened was that. I made major changes to the settings.py file and just little changes in some of the other files. I was working in another directory on my system, after the changes i just copied and replaced the same folder in this zuri-chat directory then I made a commit to same branch as before.  I was also surprised when I saw that instead of showing the new changes, it highlighted everything as “new”.

PR #163, opened 2022-11-27
Added endpoint to upload media files

babafemiolatona:
METHOD: POST

Added endpoint that allows for media upload.

args:
       org_id(str): A unique identifier of an organization.
       room_id(str): A unique identifier of the room where messages are fetched from.
       file: A file object to be uploaded.


mikengr:
This endpoint is duplicated.

mikengr:
Closed because it is Duplicated

PR #183, opened 2022-12-06
Fix sendmessages

ibimina:
Added a function that creates a random Id  to attach to the chat data (no message will have an id of undefined since this is an asynchronous function)
So the filter function in zc_main would map out the new message
Until the sent message gets fulfilled to the database


ibimina:
@JulianaSau Kindly help merge this Pull request

PR #185, opened 2022-12-07
Feature/message thread

sulaimon23:
I implemented the create channel endpoint from zuri main. so i have to implement getting room available refresh in other to display the room name on the HEAder


sulaimon23:
@JulianaSau i think you missed this

JulianaSau (approved):

PR #190, opened 2022-12-09
Fix/refactored post threads endpoint

Shopiley:
1. Refactored POST threads endpoint to provide a more detailed response json 

Before:
image

After:
image

2. Created an endpoint to GET a single message and the threads attached (to allow easier implementation of threads on the frontend).

Please review @AI-fae @mikengr


mikengr (approved):
LGTM

PR #191, opened 2022-12-09
Feat/gg 32 zuri chat filtering

bolexs:
feat(GG-32-implement-zuri-chat-filtering) [linear link](https://linear.app/team-gear/issue/GG-32/implement-zuri-chat-filtering-in-the-backend)

1. created the filtering feature at the backend to accept a value and convert the value to a stringdate
created_at__

2. Modified the sort to use 

> `"sort": { "created_at": 1}`
instead of the intial code which was `"sort": { "_id": -1}`

**Reason**: -1 starts from newest to oldest, 1 start from oldest to newest which would make the usage of pagination on the frontend better or easier to use as it is sorted from oldest to newest using created_at as a more efficective sort parameter

3. without filtering: 
without_filter

made use of raw_query to execute filtering

kindly review @mikengr @AI-fae

### DIFF
--- a/backend/endpoints/messages.py
+++ b/backend/endpoints/messages.py
@@ -102,10 +102,14 @@ async def send_message(
         message.dict(),
     )
     # instantiate the Notication's function that handles message notification
+<<<<<<< HEAD
     try:
         await notification.messages_trigger(message_obj=message)
     except Exception as e:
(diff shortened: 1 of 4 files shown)


mikengr (approved):
LGTM

PR #192, opened 2022-12-09
changed the logic for sent message

matthew-akinola:
What i did intially was to remove any character with '@' from the text, then I remembered that when a text is sent on slack, the msg also shows the tagged user, so I corrected that


matthew-akinola:
@mikengr help me review this boss!!

PR #193, opened 2022-12-09
Updated the send messages endpoint to upload files

devvspaces:
There was two endpoints that does the same thing, sending messages, but one uploaded files too. So

- I removed the other endpoint and allowed just a single send message endpoint that upload files and send messages

## Request

t2

## Response

t1


mikengr (approved):
Awesome work @devvspaces, LGTM.

PR #196, opened 2022-12-10
corrected get messages endpoint and paginator.py to return an empty a…

Shopiley:
Problem 1:
GET messages endpoint was returning ‘internal server error’ for rooms with no messages. 

Fix:
I fixed this in the paginator.py file by putting a check for if messages is None before going on with the pagination logic.

Problem 2:
After the first fix, the endpoint would return a “detail: room does not exist or no message found” message for a room with no messages.
B87782B5-99AD-471C-95E2-E7DCFB0EF693

Fix:
I changed this to return the same data structure as when a message exists, but with the data array empty as seen below.

A425746C-B6DE-440C-B13E-8F0A8B6D5FFD

### DIFF
--- a/backend/endpoints/messages.py
+++ b/backend/endpoints/messages.py
@@ -308,12 +308,6 @@ async def get_messages(
         endpoint=f"/api/v1/org/{org_id}/rooms/{room_id}/messages",
     )
 
-    if response == []:
-        raise HTTPException(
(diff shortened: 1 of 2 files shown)


mikengr (commented):
LGTM

mikengr (approved):

PR #199, opened 2022-12-12
Thread Page Fix For Frontend and Backend

Harrylever:


frontend/zc_messaging/src/pages/threads/threads.utils.js
```diff
@@ -12,6 +12,7 @@ const getThreadHandler = async (orgId, senderId) => {
   try {
     if (orgId && senderId) {
       const getThreadRequest = await axios.get(
+        // `${BASE_URL}/org/${orgId}/member/${senderId}/threads`
```
JulianaSau, line 15: Remove this commented out code

frontend/zc_messaging/src/component/ThreadList/index.jsx
```diff
@@ -1,13 +1,13 @@
...
-import styles from "./threadList.module.css"
+import { Container } from "./threadList.style"
 
 const ThreadList = ({ threadListData }) => (
-  <div className={styles.threadListWrap}>
-    {threadListData.map(thread => (
-      <ThreadItem key={thread.message_id} Thread={thread} />
+  <Container>
+    {threadListData.map((message, index) => (
+      <ThreadItem key={index} Thread={message} />
```
JulianaSau, line 8: Use a better key. It's not advisable to use the index from the array directly like that. Make it more unique to the thread items

frontend/zc_messaging/src/component/ThreadList/ThreadItem/index.jsx
```diff
@@ -1,25 +1,36 @@
...
         <p>{richUiData.blocks[0].text}</p>
       </div>
-      <div className={styles.emoji_cont_54x4}>
+      <div className="emoji_cont_54x4">
         <p>{emojis}</p>
       </div>
-    </div>
+      {threads.map((thread, index) => (
+        <div
+          key={index.toString()}
```
JulianaSau, line 24: Same thing here for this index

JulianaSau (changes requested):

JulianaSau:
@mykie88 Please help to review the backend code in this PR

frontend/zc_messaging/src/component/ThreadList/ThreadItem/index.jsx
Harrylever: Honestly, there's no key of the thread object that has a unique attribute value to fit in as the key. But I'll see if I still change it.

frontend/zc_messaging/src/component/ThreadList/index.jsx
Harrylever: Okay

backend/endpoints/threads.py
```diff
@@ -235,7 +235,7 @@ async def get_threads(org_id: str, member_id: str, page: int = 1, size: int = 15
         room_ids.append(room['_id'])
 
     for room_id in room_ids:
-        single_room_resp = await get_room_messages(org_id, room_id, page, size)
+        single_room_resp = await get_room_messages_for_threads(org_id, room_id)
```
mikengr, line 238: `get_room_threaded_messages` is a better name for this function and the variable should `threaded_messages`

Also I don't think this your for loop is doing the correct logic as intended

backend/endpoints/threads.py
```diff
@@ -235,7 +235,7 @@ async def get_threads(org_id: str, member_id: str, page: int = 1, size: int = 15
...
         if single_room_resp == []:
             # Room not found
             pass
```
mikengr, line 241: this if statement should be joined with the next one

frontend/zc_messaging/src/component/ThreadList/ThreadItem/index.jsx
JulianaSau: You could simply use the thread ID that comes with the messages.

backend/utils/message_utils.py
```diff
@@ -103,6 +103,59 @@ async def get_room_messages(
     return response
 
 
+async def get_room_messages_for_threads(
```
mikengr, line 106: this function will just return the same result as `get_room_messages` excluding pagination. If you want to write a utility function that returns threaded messages, then the logic should be different from that go get room messages.

backend/utils/message_utils.py
```diff
@@ -81,7 +81,7 @@ async def get_room_messages(
     DB = DataStorage(org_id)
 
     skip = await off_set(page, size)
-    options = {"limit": size, "skip": skip, "sort": { "created_at": -1}}
+    options = {"limit": size, "skip": skip, "sort": { "created_at": 1}}
```
mikengr, line 84: is there a reason for modifying the sort option?

mikengr (changes requested):
@Harrylever I left some comments
Also, it seems you have been deactivated, what happened?

PR #200, opened 2022-12-13
fix: datetime value bug in message request and refactor threads logic

mikengr:
* Assigned the `datetime.utcnow()` to the `created_at` field of the message object  before `db.write` is called rather than using the value from the request object

* Refactored the post thread messages endpoint with proper functions and variable naming. Also updated the `add_to_threads_list` function to insert threads at the beginning of the list rather than appending at the end.


mikengr (commented):
LGTM

PR #204, opened 2022-12-14
Fixed File upload

devvspaces:
- File upload now sends files properly, getting back the extension as a reply
- Users can now send any kind of files, the extension returned can always be used to identify the kind of file uploaded


mikengr (approved):
LGTM

PR #205, opened 2022-12-14
fix: get message route, removed duplicate endpoint url

mikengr:
* Renamed file details endpoint url to prevent conflict with get message endpoint 
* Updated `get_single_message` endpoint to `get_message_by_id`


mikengr (commented):
LGTM

PR #206, opened 2022-12-15
Fix: channel switching issue and loading state fix

sulaimon23:


mikengr:
@billmal071 please help review this boss

sulaimon23:
@JulianaSau

frontend/zc_messaging/src/pages/message-board/index.jsx
```diff
@@ -365,14 +375,14 @@ const MessagingBoard = () => {
...
-              messages={roomChats || []}
+              isLoadingMessages={isFetching}
+              messages={isFetching ? [] : roomChats || []}
               onSendMessage={sendMessageHandler}
               onReact={reactHandler}
               onSendAttachedFile={SendAttachedFileHandler}
               currentUserId={authUser?.user_id}
               height={"92vh"}
-              onHandleScroll={handleScroll}
+              onHandleScroll={event => handleScroll(event)}
```
JulianaSau: I'm guessing this is very dependent on the PR that you made to zc main. Why is adding this event necessary. Was it not working before without it?

frontend/zc_messaging/src/pages/message-board/index.jsx
```diff
@@ -365,14 +375,14 @@ const MessagingBoard = () => {
           <MessageWrapper>
             <MessageRoomViewHeader name={`#${roomName}`} />
             <MessageBoard
-              isLoadingMessages={isLoadingRoomMessages}
-              messages={roomChats || []}
+              isLoadingMessages={isFetching}
```
JulianaSau, line 378: I think adding both is fetching and is loading would be more beneficial. Because `isLoadingRoomMessages` only occurs when the component is mounted for the first time. Then the `isFetching`happens for all subsequent fetches that happen during the session when one is active on the application when its listening to any changes that occur when the data is updated

JulianaSau (changes requested):

frontend/zc_messaging/src/pages/message-board/index.jsx
sulaimon23: The handle scroll function was emitting without it being called. Found a fix. so i have removed the event

frontend/zc_messaging/src/pages/message-board/index.jsx
sulaimon23: The isLoadingRoomMessages trigger on the first load of the components. it doesn't trigger on channel change or on subsequent changes which makes it look as though it is not working

JulianaSau: Okay. I just re-read the docs and realised they updated the distinction between the two. Which is fine.
I guess isFetching can work when it's alone in this scenario

sulaimon23: OKay. Thanks

sulaimon23:
@ickynavigator

sulaimon23:
@billmal071

JulianaSau (approved):

PR #208, opened 2022-12-25
feat: added endpoint to update a thread message

Shopiley:
Implemented update threads endpoint.

Note:
This is a temporary solution as it uses an O(n) algorithm while we review a better way to implement threads (possibly as a collection of it's own).
Also when testing the PUT /Update Message endpoint, the thread_id must be included in the request body.

@mikengr @zxenonx @AI-fae

### DIFF
--- a/backend/endpoints/threads.py
+++ b/backend/endpoints/threads.py
@@ -5,9 +5,10 @@ from schema.response import ResponseModel
 from schema.thread_response import ThreadResponse
 from fastapi.responses import JSONResponse
 from utils.centrifugo import Events, centrifugo_client
-from utils.message_utils import get_message, get_room_messages, update_message_threads as edit_message_threads
+from utils.message_utils import get_message, get_room_messages
 from utils.room_utils import get_org_rooms
 from utils.threads_utils import get_message_threads, add_message_to_thread_list
+from utils.threads_utils import  update_thread_message as edit_thread_message
 import json
 
 router = APIRouter()
@@ -43,18 +44,33 @@ async def send_thread_message(
             A dict containing data about the message that was edited.
 
             {
-                "status": 201,
-                "event": "thread_message_create",
-                "thread_id": "bd830644-2205-11ec-9853-2ff0a732e3ef",
-                "room_id": "614e1606f31a74e068e4d2e2",
-                "message_id": "6155a0e6be7f31a9275a1eca",
-                "data": {
-                    "sender_id": "61467e181ab13c00bcc15607",
-                    "message": "Checking out the threads",
-                    "created_at": "2021-09-30T15:41:45.685000Z"
-                }
-            }   
-
+            "status": 201,
+            "event": "thread_message_create",
+            "room_id": "637f6f2d601ce3fc5dc738f5",
+            "message_id": "63a6c075d4093d73264466c6",
+            "data": {
+                "sender_id": "637f6f28601ce3fc5dc738f4",
+                "emojis": [],
+                "richUiData": {
+                "blocks": [
+                    {
+                    "key": "f3s6p",
+                    "text": "ho ho ho I'm a thread message on christmas eve!",
+                    "type": "unstyled",
+                    "depth": 0,
+                    "inlineStyleRanges": [],
+                    "entityRanges": [],
+                    "data": {}
+                    }
+                ],
+                "entityMap": {}
+                },
+                "timestamp": 1669917458839,
+                "thread_id": "75d6507d-836b-11ed-9fba-40234355d0d0",
+                "created_at": "2022-12-24 09:15:08.019726"
+            }
+            }
+            
         Raises:
             HTTPException [401]: You are not authorized to edit this message.
             HTTPException [404]: Message not found.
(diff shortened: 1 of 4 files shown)


backend/endpoints/threads.py
```diff
@@ -5,9 +5,10 @@
 from schema.thread_response import ThreadResponse
 from fastapi.responses import JSONResponse
 from utils.centrifugo import Events, centrifugo_client
-from utils.message_utils import get_message, get_room_messages, update_message_threads as edit_message_threads
+from utils.message_utils import get_message, get_room_messages
 from utils.room_utils import get_org_rooms
 from utils.threads_utils import get_message_threads, add_message_to_thread_list
+from utils.threads_utils import  update_thread_message as edit_thread_message
```
FlorenceEgwu, line 11: why is the alias neccessary?

backend/endpoints/threads.py
```diff
@@ -168,42 +184,33 @@ async def update_threads_message(
     org_id: str,
     room_id: str,
     message_id: str,
+    thread_id: str,
```
FlorenceEgwu, line 187: why not provide this as a path parameter for the purpose of consistence?

backend/requirements.txt
```diff
@@ -14,7 +14,6 @@ colorama==0.4.4
 coverage==6.0.2
 distlib==0.3.3
 fastapi==0.70.0
-fastapi-pagination==0.9.1
```
FlorenceEgwu, line 17: how is this related to the current changes?

backend/requirements.txt
```diff
@@ -46,7 +45,6 @@ python-dateutil==2.8.2
 python-dotenv==0.19.1
 python-multipart==0.0.5 
 PyYAML==6.0
-regex==2021.10.23
```
FlorenceEgwu, line 49: why???

backend/utils/threads_utils.py
```diff
@@ -56,8 +56,40 @@ async def add_message_to_thread_list(org_id, room_id, message_id, request: Threa
...
+         payload (dict[str, Any]): The new data.
+         message (dict[str, Any]): The message to be edited that has been loaded from the database
+     Returns:
+         dict[str, Any]: The response returned by DataStorage's update method.
+     """
+    threads = message["threads"]  # a list of thread objects
+
+    # to edit a thread object in the list of threads
+    for index in range(len(threads)):
+
```
FlorenceEgwu, line 83: Like you already mentioned in your PR description, this is a potential problem. we can't be looping through 100 threads just to update 1

backend/utils/threads_utils.py
```diff
@@ -56,8 +56,40 @@ async def add_message_to_thread_list(org_id, room_id, message_id, request: Threa
...
+            threads[index]["edited"] = True
+
+            del message["_id"]
+
+            db = DataStorage(org_id)
+            return await db.update(settings.MESSAGE_COLLECTION, message_id, message)
+
+    raise HTTPException(
+        status_code=status.HTTP_404_NOT_FOUND, detail="Thread message not found or sender is not authorised to edit this message"
+    )
```
FlorenceEgwu, line 95: I think it will be best if 2 different exceptions are raised here. One it a 404 exception while the other is a 403 (where the sender is not the same) exception.
Best to keep things explicit for readability in the near future.

FlorenceEgwu (changes requested):

FlorenceEgwu:
Good Job Sope!
I have left a few comments.

BTW, is there anything particularly stopping us from restructuring things and implementing this threads feature correctly right away?

backend/endpoints/threads.py
zxenonx: @Shopiley, please rename the endpoint function to **update_thread_message** and that would make sense.

zxenonx, line 11: Kindly merge these imports.

backend/endpoints/threads.py
zxenonx: Yes please, the id of the resource to be edited must be provided in the route. @Shopiley

zxenonx: Also please add a docstring to the function.

backend/endpoints/threads.py
```diff
@@ -168,42 +184,33 @@ async def update_threads_message(
...
-    if not added_thread_message:
-        raise HTTPException(
-            status_code=status.HTTP_424_FAILED_DEPENDENCY,
-            detail={"thread not added": added_thread_message},
-        )
+    payload = request.dict(exclude_unset=True)
 
-    # payload["edited"] = True
-    loadedMessage.update(thread_to_add)
+    edited_thread_message = await edit_thread_message(org_id, message_id, thread_id, payload, message)
```
zxenonx, line 201: This variable is unused...
You should check whether the thread msg has been successfully edited or not.

backend/endpoints/threads.py
```diff
@@ -168,42 +184,33 @@ async def update_threads_message(
...
         room_id,
         Events.MESSAGE_UPDATE,
-        loadedMessage
+        message
     )
 
     return JSONResponse(
         content=ResponseModel.success(
-            data=loadedMessage, message="Thread Added"),
+            data=message, message="Thread Edited"),
```
zxenonx, line 213: Thanks for the correction here.
loadedMessage is not a valid variable in Python.

backend/utils/message_utils.py
```diff
@@ -179,18 +179,19 @@ async def update_message(
...
-        message_id (str): The id of the message to be edited.
-        message (dict[str, Any]): The new data.
-    Returns:
-        dict[str, Any]: The response returned by DataStorage's update method.
-    """
-
-    db = DataStorage(org_id)
-
-    return await db.update(settings.MESSAGE_COLLECTION, message_id, message)
+# async def update_thread_message(
```
zxenonx, line 182: If it will no longer be used, we can just delete it.

backend/utils/threads_utils.py
```diff
@@ -56,8 +56,40 @@ async def add_message_to_thread_list(org_id, room_id, message_id, request: Threa
...
-   
+
     response = await update_message(org_id, message_id, message)
-    
+
     return response, thread_message
 
+
+async def update_thread_message(
+    org_id: str, message_id: str, thread_id: str, payload: dict[str, Any], message: dict[str, Any]
```
zxenonx, line 66: Do we still need **message** when **message_id** is already provided?
5 params, This is starting to be a lot.

backend/utils/threads_utils.py
zxenonx: We need to find a better implementation here.
But just a quick suggestion, in this situation it's better to use **enumerate** instead of **range+len**.

https://themeptation.medium.com/why-use-enumerate-instead-of-range-in-your-pythons-loops-d33bfd9c7531

zxenonx (changes requested):
Good job @Shopiley.
Kindly go through the comments I left.
Thanks.

Shopiley:
Thanks for comments! I've made [another PR](https://github.com/zurichat/zc_messaging/pull/213) implementing an improved solution. I also made corrections based on your comments here.

backend/utils/threads_utils.py
Shopiley: Done

PR #213, opened 2023-01-31
Feat/update threads endpoint

Shopiley:
Implemented update threads endpoint. An improved solution from [Previous PR](https://github.com/zurichat/zc_messaging/pull/208)

pls review


backend/utils/db.py
```diff
@@ -166,12 +171,18 @@ async def update(
             RequestException: Unable to connect to zc_core
         """
 
+        # to ensure only either one of raw_query or data is sent
+        if data and raw_query:
+            raise Exception("Either one of data or raw_query is expected")
```
mikengr, line 176: Is there a reason this conditional is needed? what happens when we have both data and raw_query?

backend/endpoints/threads.py
```diff
@@ -164,96 +158,83 @@ async def get_thread_messages(org_id: str, room_id: str, message_id: str):
...
+    """
 
-    # Publish to centrifugo in the background.
-    background_tasks.add_task(
-        centrifugo_client.publish,
-        room_id,
-        Events.MESSAGE_UPDATE,
-        loadedMessage
-    )
+    message = await get_message(org_id, room_id, message_id)
```
mikengr: why do we need to get the message here? we are making another db call in `threads_utils` to get the same message.

backend/.pre-commit-config.yaml
```diff
@@ -14,7 +14,7 @@ repos:
       - id: pretty-format-json
 
   - repo: https://github.com/psf/black
-    rev: 21.9b0
+    rev: 22.3.0
     hooks:
```
mikengr, line 18: how about this? will something break if we merge this? I can see it is a config file.

mikengr (commented):
Awesome work @Shopiley, I left some comments.

backend/.pre-commit-config.yaml
```diff
@@ -45,3 +45,7 @@ repos:
           - --disable=R0902
           - --disable=R0903
           - --disable=C0103
+          - --disable=R0913
+          - --disable=C0327
+          - --disable=C0305
```
o4codes: Why were some pre-commit hooks disabled ?

backend/endpoints/threads.py
Shopiley: oh right, I could just return the message from the db call we made in `threads_utils` instead then

backend/.pre-commit-config.yaml
Shopiley: Yeah I just updated the version of black to the latest version to take care of some of the issues with I was experiencing with the old version following a suggestion by the author on stackoverflow.

backend/.pre-commit-config.yaml
Shopiley: I disabled the `RO913: Too many arguments (6/5)` hook because it was complaining about the update( ) function in db.py (of which the 6 arguments it has are all very much needed)

but as for the rest, I'll enable them - my bad

backend/utils/db.py
Shopiley: image
this code snippet is from [ zc_core's `handlePut` function](https://github.com/zurichat/zc_core/blob/dev/data/write_data.go)

Because of this if-else block, when both data and raw_query are passed, zc_core will always choose to update with the raw_query. So if you intend to update a message for instance, you should only pass in data (a.k.a payload). But if you intend to update a thread message and you pass in both, zc_core will only update with raw_query anyways.

Shopiley: By the way I've improved the if statement because I realised it wasn't accounting for a situation where both data and raw_query aren't provided.

mikengr: Okay sounds good.

backend/utils/threads_utils.py
```diff
@@ -56,8 +56,53 @@ async def add_message_to_thread_list(org_id, room_id, message_id, request: Threa
...
+        dict[str, Any]: The response returned by DataStorage's update method.
+    """
+    message = await get_message(org_id, room_id, message_id)
+
+    if not message:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND, detail="Message not found"
+        )
+
+    if message["sender_id"] != payload["sender_id"]:
```
mikengr: @Shopiley I think we have an implementation problem on this line. The sender of a parent message is not necessarily the author of a thread message in the threads list. We need to be checking `thread_message["sender_id"]` here rather than the `message["sender_id"]`.

mikengr (changes requested):
@Shopiley I left one more comment.

backend/utils/threads_utils.py
Shopiley: Oh that's true, but once again the issue of not being able to access the thread_id directly pops up

mikengr: we need to write a query that retrieves the thread message itself

Shopiley: Done, pls review.

backend/utils/threads_utils.py
```diff
@@ -28,6 +28,39 @@ async def get_message_threads(org_id, room_id, message_id):
     message = await get_message(org_id, room_id, message_id)
     return message["threads"]
 
+async def get_message_thread(org_id, room_id, message_id, thread_id):
+    """Retrives a single the messages in a thread.
+
```
FlorenceEgwu, line 33: Did you mean "Retrieves a single thread message"?

backend/utils/threads_utils.py
```diff
@@ -28,6 +28,39 @@ async def get_message_threads(org_id, room_id, message_id):
...
+        collection_name=settings.MESSAGE_COLLECTION,
+        query=query,
+        options=options
+    )
+    
+    if not response or "status_code" in response:
+        return []
+
+    return response["threads"]
+
```
FlorenceEgwu, line 62: I am struggling to understand what this function does exactly.
Does it retrieve a single thread from a list of a message's threads or all the threads under that message?

I ask because, if it is a single thread message, then this should probably return an object (dictionary) instead of a list.
will appreciate a little bit of clarification

FlorenceEgwu:
You've done a great job here @Shopiley .
I left a few questions for clarification.

backend/utils/threads_utils.py
Shopiley: yes, my bad. I've corrected it.

backend/utils/threads_utils.py
Shopiley: Yes, it retrieves a single thread from the list of threads in a message. But because the `threads` field in the message object is a **list** of thread objects, a query to zc_core would return a **list** as well (same data type) of a single thread (by thread_id specified). 

But as per your suggestion to make the function return an object instead, I could change it to return `response["threads"][0]` .

FlorenceEgwu: okay, that makes sense.

zxenonx (approved):
Good job @Shopiley

